PR #2, opened 2021-09-20
✨Add i18n chinese translation

canstand:
Add i18n chinese translation, and some slightly modify


jpanther:
Thanks for your contribution, @canstand.

PR #10, opened 2021-10-17
Copyright and Footer Improvements

baalajimaestro:
* Added support for pre-footer partial which can enable users to add sitelinks to all sites, above the copyright
* Also enabled year/copyright symbol even when site.copyright is set.

### DIFF
--- a/layouts/partials/footer.html
+++ b/layouts/partials/footer.html
@@ -2,6 +2,8 @@
   {{/* Copyright */}}
   <p class="text-neutral-400 dark:text-neutral-500">
     {{- with .Site.Copyright }}
+      &copy;
+      {{ now.Format "2006" }}
       {{ . | emojify | markdownify }}
     {{- else }}
       &copy;
--- a/layouts/partials/footer.html
+++ b/layouts/partials/footer.html
@@ -1,4 +1,8 @@
 <footer class="py-10 text-sm">
+  {{/* Footer Pre */}}
+  {{ if templates.Exists "partials/pre-footer.html" }}
+    {{ partialCached "pre-footer.html" . }}
+  {{ end }}
   {{/* Copyright */}}
   <p class="text-neutral-400 dark:text-neutral-500">
     {{- with .Site.Copyright }}


jpanther:
Thanks for these suggestions. However, I have some concerns about these changes.

Firstly, the copyright message is intended to be customisable. I already know of sites that are using Congo that do not include the copyright symbol or year in their footer. I have seen plenty of other Hugo templates where site owners use other copyright messages such as "CC-BY-NA" in place of the copyright symbol. I think this is too restrictive a change and forces users to adopt a particular style that may not even be compatible with languages other than English. There's nothing to stop people including the copyright symbol and year in their custom message either (albeit without the variable for the year).

With regards to the pre-footer. I can see there is some value in allowing for additional content to be inserted into the base template. I'm concerned that this is probably a change that needs a bit more thought. Part of me leans towards it probably being easier to allow users to override the footer layout whenever they want to make a holistic change like this. Perhaps a cleaner solution, therefore, is adjusting the footer layout to allow it to be more readily overridden without breaking core functionality. Otherwise the risk is we will end up with all these optional layout files that get inserted and it becomes more difficult for people to understand.

baalajimaestro:
I see the reason, and yes it makes total sense to me.
Shall I redo it, adding a config to enable it if needed? 

Maybe like `site.copyrightdate=true` and `site.copyrightsymbol="anything that they want"` for example, if someone wants to add &reg; or &trade; and still wants to keep using the autogenerated year, it would make this much easier. `site.copyrighttext` to hold the text on this case.

site.copyright would be fully overriding the copyright section with the text specified.

Lemme know if this is fine and i will push it

Do you mean overriding the entire footer on this case?

jpanther:
Adding extra config fields to the copyright was my first thought with this too but I'm not sure that's a very clean solution either. I'm actually wondering if this is better solved using i18n. We could define a localisable string for the copyright line and then provide the year, title and user-defined string as a parameters.

For example the default i18n string could be: `"&copy; {{ .Year }} {{ .Title }}"` but users could override this to be `"&copy; {{ .Year }} {{ .Copyright }}"`

This way the string is localisable and can be completely overridden by users without having to set all these different configuration variables.

On the topic of footer links, I've pushed some changes to the dev branch which adds an optional menu to the footer. It's configured the same way as the main menu. This enables links to be added easily without having to define a partial. Does this help to address your use case?

baalajimaestro:
Closing this, the footer change that was added, is just right for my need. Thanks a lot!

PR #14, opened 2021-10-20
:memo: Fix contribution guide URL; add update congo info

seahagcafe:
Hi! Like other contributors here, I'm using your template for my own site (to be deployed). I'm new to contributing, so please let me know if I need to change anything with how I'm doing this.

Thanks,
M

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,6 +113,12 @@ links = [
 ]
 ```
 
+## Keep Congo Updated
+
+Run `go get github.com/jpanther/congo@upgrade` to get Congo's latest version.
+
+If you run into the error starting with `go mod download: github.com/jpanther/congo@v1.3.0: invalid version`, then you'll need to get the latest version.
+
 ## Advanced customisation
 
 Refer to the [theme documentation](https://jpanther.github.io/congo/docs/advanced-customisation/) for details on customising Congo, including rebuilding the theme from scratch.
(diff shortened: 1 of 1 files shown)


jpanther:
Hey @myriahnottage, thanks for these suggestions. The URL change is fine (thanks for finding it!) but the update instructions probably need to cater to a broader audience. The issue is that people can install the theme many different ways and these instructions assume they have installed it using Hugo Modules.

I usually use the command `hugo mod get -u` which will automatically find the latest version of any installed modules. Is there a particular reason you use the `go` command instead? I wonder if this is why you were seeing the error message about the invalid version.

_And on the contributing... it looks like you created a pull request on your fork which is not something you generally need to do (that's why there's two commits appearing above). For this particular project too, you should also use Gitmoji at the start of each commit message._

seahagcafe:
> Hey @myriahnottage, thanks for these suggestions. The URL change is fine (thanks for finding it!) but the update instructions probably need to cater to a broader audience. The issue is that people can install the theme many different ways and these instructions assume they have installed it using Hugo Modules.
> 
> I usually use the command `hugo mod get -u` which will automatically find the latest version of any installed modules. Is there a particular reason you use the `go` command instead? I wonder if this is why you were seeing the error message about the invalid version.
> 
> _And on the contributing... it looks like you created a pull request on your fork which is not something you generally need to do (that's why there's two commits appearing above). For this particular project too, you should also use Gitmoji at the start of each commit message._

Ooh that makes sense, I was definitely struggling to get hugo to run today, but it looks like I forgot to NOT commit the .DS_Store files that can appear from macs >.<
At any rate that particular issue seems to be resolved for now. I'll investigate a bit more tomorrow and possibly add a call out, though mac users in general may generally already be in the know about that.

Just to clarify on gitmoji -- I see now that others added them to actual commit messages, not **just** the PR description. Thanks for the reminder! I'll go ahead and close this out to get those gitmojis in there!

jpanther:
Ah yes, the .DS_Store files are annoying (I'm a Mac user too). A tip is to add them to your global .gitignore file that lives at `~/.gitignore`. That way you'll never accidentally check them in. If you check out [gitignore.io](https://gitignore.io) you can see some common templates that can be used to block all the common files for macOS.

PR #26, opened 2021-11-06
Fix spacing for pagination dates

Benjamin-Lee:
Before and after:

image

With the current styling, I think the spacing is too cramped in the pagination. This PR uses the default margin instead.

### DIFF
--- a/assets/css/compiled/main.css
+++ b/assets/css/compiled/main.css
@@ -1928,8 +1928,8 @@ body a, body button {
   margin-right: 0.75rem;
 }
 
-.-mt-1 {
-  margin-top: -0.25rem;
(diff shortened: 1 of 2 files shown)


jpanther:
Thanks for the PR, @Benjamin-Lee. I think this change is good but might still need a little tweak. I imagine when I styled this I didn't try it out with such long titles and the line height still looks off when it wraps onto multiple lines. I wonder if adding something like `leading-5` or `leading-6` would help to tighten it up a bit?

A quick mock-up before and after...

Screen Shot 2021-11-06 at 13 38 30

Benjamin-Lee:
Yes, much better with the leading fix. Thanks!

PR #27, opened 2021-11-06
Adding German localization (de i18n)

max-alletsee:


jpanther:
Thanks! I'll merge these in.

PR #30, opened 2021-11-10
:globe_with_meridians: Add es i18n

iyanmv:
Add Spanish (Spain) translation

### DIFF
--- a/i18n/es.yaml
+++ b/i18n/es.yaml
@@ -1,0 +1,37 @@
+article:
+  anchor_label: "Ancla"
+  draft: "Borrador"
+  edit_title: "Editar contenido"
+  reading_time:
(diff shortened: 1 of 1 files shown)


jpanther:
Thanks for your PR. The commit message action is having a few issues that I'm working on at the moment so don't worry about that. I'll merge your change in. :)

iyanmv:
> Use Gitmoji in commit messages to provide context.

What a pain to comply with this... The ascii code from https://gitmoji.dev/ is not okay, copying-pasting the utf8 is not okay either...

iyanmv:
Thanks! I gave it a try... but it's too late to deal with emojis hahaha :see_no_evil:

PR #32, opened 2021-11-16
Fix pagination uses the date of current article; update zh i18n

canstand:
Pagination uses the date of the current article and should use the date of the specified article.


jpanther:
Thanks for these changes, @canstand. I'll merge them in.

PR #37, opened 2021-11-22
:memo: add user

AntoineSoetewey:
Add antoinesoetewey.com to the list of congo users


jpanther:
Thanks for using Congo! I'll merge this in.

PR #54, opened 2021-12-16
Add word counts in post meta.

wizardbyron:


wizardbyron:
I committed with wrong email.

PR #57, opened 2021-12-16
Add word counts in page meta.

wizardbyron:


jpanther:
Thanks for the PR. I have one suggestion here... it would be preferable to pass `.WordCount` to the i18n function (similar to how the reading time works) so that the string can be translated more easily. It would then allow supporting plurals and there may be instances where some languages need to prefix the counter.

**Example:**

```go
{{- i18n "article.word_count" .WordCount | emojify -}}
```

```yaml
article:
  word_count:
    one: "{{ .Count }} word"
    other: "{{ .Count }} words"
```

jpanther:
Thanks! I'll merge this in.

PR #61, opened 2021-12-24
Add dr460nf1r3.org to users

dr460nf1r3:


jpanther:
Great looking site! Thanks for using Congo. I'll merge this in and it will appear in the public list when the next release of the theme is published.

PR #65, opened 2021-12-25
Fix error unclosed action in footer.html

darkrain-nl:
Fix error unclosed action

"Error: add site dependencies: load resources: loading templates: "/tmp/hugo_cache/modules/filecache/modules/pkg/mod/github.com/jpanther/congo@v1.6.0/layouts/partials/footer.html:35:1": parse failed: template: partials/footer.html:35: unclosed action"

### DIFF
--- a/layouts/partials/footer.html
+++ b/layouts/partials/footer.html
@@ -32,8 +32,7 @@
       {{ if .Site.Params.attribution | default true }}
         <p class="text-xs text-neutral-400 dark:text-neutral-600">
           {{ $hugo := printf `<a class="hover:underline hover:underline-primary-400 hover:text-primary-500"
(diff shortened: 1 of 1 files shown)


jpanther:
Thanks for the PR. I'm surprised at this error as this particular line of code has been present for some time. What version of Hugo are you using?

darkrain-nl:
> Thanks for the PR. I'm surprised at this error as this particular line of code has been present for some time. What version of Hugo are you using?

That was the stable package in Debian 11, version 0.80.0-6+b5..
I know that was an old version and not really supported.

But because it seems an easy fix, I thought to do a PR.

jpanther:
Ah that makes sense then. The theme is only supported on Hugo v0.86 and later so that's probably why this hasn't come up previously. In any case, I'm happy to accept the fix.

PR #66, opened 2021-12-27
Add parentheses in date.html

AntoineSoetewey:
When following this [workflow](https://bookdown.org/yihui/blogdown/workflow.html), there is an error when serving the site :

Screenshot 2021-12-27 at 11 55 18

By adding these parentheses, the site is served without any error.

Regards,
Antoine


jpanther:
Thanks for this, happy to merge it in.

PR #76, opened 2022-01-05
adapt year in footer

AntoineSoetewey:
I have seen from another site that this script adapts the year automatically without the need to serve the site each year.

I’m not an expert, so I’ll let you decide whether this is useful or not.

### DIFF
--- a/layouts/partials/footer.html
+++ b/layouts/partials/footer.html
@@ -24,7 +24,7 @@
           {{ . | emojify | markdownify }}
         {{- else }}
           &copy;
-          {{ now.Format "2006" }}
+          <script>document.write(new Date().getFullYear())</script>
           {{ .Site.Author.name | markdownify | emojify }}
         {{- end }}
       </p>


jpanther:
While this does work, I want to move towards removing any inline scripting and this change goes against that objective - see https://github.com/jpanther/congo/discussions/68.

AntoineSoetewey:
Thanks for your reply.

Just to make sure I understand correctly (sorry if the question is trivial, I am not an expert), this script may compromise security and safety of the website?

jpanther:
The question certainly isn't trivial... and neither is the answer really. This particular change probably isn't a high risk and inline Javascript on the whole isn't 100% bad however there are many considerations. As this is just a one-liner script it's not a great example, but if you can imagine the overall website and consider the trade-offs. One of the main concerns is Javascript can be exploited using many different methods and one of the defences against this is to implement security response headers on your web server.

While there are now quite a few of these headers available, one that is becoming more common is 'Content Security Policy' (CSP) which allows site owners to specify where certain types of assets are permitted to be loaded from. As part of this you can specify that you do not want to permit inline scripting. This has the benefit of preventing a number of possible exploit vectors but unfortunately it also has the side effect of preventing legitimate inline Javascript from being able to run.

Now this is where it's a bit of a trade-off. It's definitely beneficial to be able to set a CSP header and currently the theme doesn't make this easy. The trade-off is that this creates design constraints around how Javascript can be implemented within the theme. You may find it interesting to [read more about CSP](https://developers.google.com/web/fundamentals/security/csp/) here. So essentially the change you've proposed isn't in itself risky, but it's more that the architecture of having the inline scripting in the page to begin with is a risk.

I should say this is only one reason for why you might choose to handle inline scripting a certain way and I'm giving it because it's the main reason behind why I'm trying to avoid it.

AntoineSoetewey:
Thanks for this complete answer, it makes much more sense now and I’ve learned something!

Issue #83, opened 2022-01-18
No emojify on list page


rshmhrj:
### What happened?

Titles are `emojified` in `single.html`:
![image](https://user-images.githubusercontent.com/750874/150025613-af6a39d2-043d-4ab6-8969-4e78ac5b7250.png)

but not in `list.html`:
![image](https://user-images.githubusercontent.com/750874/150025666-79c475d3-96f3-4111-997e-f1d9c91f4e1e.png)

Propose updating `layouts/partials/article-link.html` from `>{{ $.Title }}</a` to `>{{ $.Title | emojify }}</a`:

Result:
![image](https://user-images.githubusercontent.com/750874/150025818-22012e90-9fab-4e20-a211-49df55d1b610.png)

### Theme version

v1.6.2

### Hugo version

v0.91.1

### What browsers are you seeing the problem on?

Chrome

rshmhrj:
Also suggest updating `head.html` as well:
`<title>{{ .Title }} &middot; {{ .Site.Title }}</title>`

![image](https://user-images.githubusercontent.com/750874/150026725-2a0bc478-5f6e-4cb2-8db5-1fd5cd932fab.png)


to 
`<title>{{ .Title | emojify }} &middot; {{ .Site.Title }}</title>`

![image](https://user-images.githubusercontent.com/750874/150026848-8d11416b-8fe2-4751-9679-b2dc5a7f5638.png)

jpanther:
Thanks for picking this one up, definitely a bug. I think it should probably be added to the `.Site.Title` too as it has `emojify` in the header partial but not in the HTML `title`.

rshmhrj:
@jpanther -- updated the PR.

Question: will ` | safeHTML | emojify` stack / pipe correctly?

```html
{{ printf `<link rel="%s" type="%s" href="%s" title="%s" />` .Rel .MediaType.Type .RelPermalink $.Site.Title | safeHTML | emojify }}
```

jpanther:
That one might cause problems... it may need to be something like this:

```go
{{ printf `<link rel="%s" type="%s" href="%s" title="%s" />` .Rel .MediaType.Type .RelPermalink ($.Site.Title | emojify) | safeHTML }}
```

rshmhrj:
Made the fix and pushed to the PR

PR #84, opened 2022-01-18
🐛  add emojify for title on list page and browser / tab title

rshmhrj:
Fix for #83 which transforms text to emojis, e.g.  `:book:`  to  📖 

Updates are on:
- article-link.html (effects content title in list views)  
  image

- head.html (effects browser / tab title)
  image


jpanther:
Thanks, I'll merge this fix in.

PR #91, opened 2022-01-23
💄 Adjust metadata output

cbelena:
Hi @jpanther! Per discussion #88:

I have submitted two commits:

1. Add i18n 'Updated' strings: I have added the `article.updated` string to the different languages available, making possible the translations to the last updated date in article's metadata. I've also updated the missing spanish translations of "word" and "words".
2. Adjust metadata output: I have created a new partial called `date-updated.html` which adds the 'Updated' localized string and then calls the `date.html` partial. The new simplified `date.html` also solves a little problem where before it displayed both the date and the updated date on pagination. Then, in `article-meta.html`, instead of showing the last updated date in parentheses, I have treated it like another meta parameter.

This is my first ever pull request, I'm sorry if I did something wrong :sweat_smile:

### DIFF
--- a/exampleSite/content/samples/placeholder-text.md
+++ b/exampleSite/content/samples/placeholder-text.md
@@ -1,9 +1,11 @@
 ---
 title: "Placeholder Text"
 date: "2019-03-09"
+lastmod: "2022-01-24"
 draft: true
 description: "Lorem Ipsum Dolor Si Amet"
 tags: ["markdown", "text", "sample", "latin"]
+showDateUpdated: true
 xml: false
 ---
 
--- a/i18n/de.yaml
+++ b/i18n/de.yaml
@@ -6,6 +6,7 @@ article:
     one: "{{ .Count }} min"
     other: "{{ .Count }} min"
   reading_time_title: "Lesezeit"
+  # updated: "Updated"
   # word_count:
   # one: "{{ .Count }} word"
   # other: "{{ .Count }} words"
--- a/i18n/en.yaml
+++ b/i18n/en.yaml
@@ -6,6 +6,7 @@ article:
     one: "{{ .Count }} min"
     other: "{{ .Count }} mins"
   reading_time_title: "Reading time"
+  updated: "Updated"
   word_count:
     one: "{{ .Count }} word"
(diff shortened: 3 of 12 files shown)


cbelena:
Oh, I have not included my CSS fix to wrap the text around the metadata because I'm not sure if that's the best way to do it :sweat_smile:

jpanther:
Thanks for the PR @cbelena. A very good first pull request! I'm happy with the changes however my only concern is that using the `time.Format` function requires a bump in the minimum Hugo version which prevents this from being a patch release. Can we leave this with `Date.Format` for now and it will pick up the new function in v2 of the theme?

And yes don't worry about the CSS. I'll add a commit that addresses that part for you. It requires the Tailwind CSS to be recompiled.

jpanther:
I'm also now wondering if we should pass the date into `i18n` as a variable too. I'm not sure but there may be some languages that require the format "{Date} Updated" rather than "Updated {Date}".

cbelena:
> Thanks for the PR @cbelena. A very good first pull request! I'm happy with the changes however my only concern is that using the `time.Format` function requires a bump in the minimum Hugo version which prevents this from being a patch release. Can we leave this with `Date.Format` for now and it will pick up the new function in v2 of the theme?

Sure! I reverted to `Date.Format` instead of `time.Format`.

> And yes don't worry about the CSS. I'll add a commit that addresses that part for you. It requires the Tailwind CSS to be recompiled.

Wow, that was easy!! I have to learn more CSS! :sweat_smile::sweat_smile: Thanks!

> I'm also now wondering if we should pass the date into `i18n` as a variable too. I'm not sure but there may be some languages that require the format "{Date} Updated" rather than "Updated {Date}".

I have left it the same for now because it has a colon between the 'Update' and the date, like: `Updated: Jan 24, 2022`. But I'm not sure either...

PS: Oops, I forgot to add the globe emoji to the commit that updates the TR translation, sorry!

jpanther:
The CSS turned out to be as simple as adding `flex-wrap` to the div so this was an easy fix. I think the rest looks good though so I'll merge this in now.

PR #110, opened 2022-02-08
Update de.yaml

max-alletsee:
... to add the new strings introduced since the initial translation


jpanther:
Thanks for the update. I'll merge these changes in.

PR #134, opened 2022-02-23
Adds note box ✨ 🧙

OCram85:
#### :book: Summary

Adds an additional al note box.

Files added/ changed:

- added comment.svg icon
- added notes shortcode
- added comment icon in icon samples post
- added note section in shortcote post

#### :bookmark_tabs: Test Plan

> :bulb: Select your test plan for the code changes.

- [x] Tested locally
- [x] Tested via Example page
- [x] Tested with Netlify test build 

#### :books: Additional Notes

- Already in use at [OCram85.com](https://next.ocram85.com/posts/pwsh-read-only-class-properties/)
  - Currently updating to `Congo 2`

#### A word about boxes:

I think there could be 2 approaches for multiple notification box types:

###### A generic one which could be implemented like this

There should be one `box` type shortcode with multiple params:

```
- bg-color
- icon
- prefix
- text
```` 

- 👍  reduces overall code duplication
- 👍 gives max flexibility to users how to use the boxes
- 👎 The will be not unique usage. I see a mix of icons and prefixes

###### Multiple static ones

Specific boxes for use cases. There is just the content/message as param.

- 👍 Simplified usage
- 👍 Unique look and feel
- 👎 Less flexibility for users

### CODE REVIEW

File: layouts/shortcodes/note.html [hunk 1]
```diff
@@ -0,0 +1,8 @@
+<div class="flex px-4 py-3 rounded-md bg-primary-100 dark:bg-neutral-700">
+  <span class="ltr:pr-3 rtl:pl-3 text-primary-400">
+    {{ partial "icon.html" "comment" }}
+  </span>
+  <span class="no-prose dark:text-neutral-300">
+    <strong>NOTE: </strong> {{- .Inner | markdownify -}}
```
fteem: 👋 Thank you for adding this, I was looking for such a shortcode while porting my blog to Congo.

Could this new shortcode follow the same approach as [alert](https://github.com/jpanther/congo/blob/2ceab5d4b0106549a77edcf86d4c40f301282893/layouts/shortcodes/alert.html), i.e. do not hard-code `NOTE`? 🤔

OCram85: Sure I can remove the `Note` prefix.

Just want to share on thought:
 - 👍  I like to add as much content as possible to the shortcode.
   - If you use the note/alert boxes you could end up with `ALERT: / Alert: | NOTE: / Note: ` in multiple variations
 - 👎 With the included prefix it takes the possibility for custom prefix patterns.

On either way, alert and note should be identical in this behavior.

fteem: I believe that hard-coded shortcodes are also hard/impossible to i18n, and i18n is a first-class citizen in Congo.

OCram85: That's a good point. I have not thought about that

OCram85: So I suggest the following:
- I remove the hard coded prefix for this PR.
- I'm going to bring up another PR which includes the prefix for note, alert and other shortcodes. These will be using i18n.

What do you think?

@jpanther: Would this also work for you?

fteem: Thanks for thinking it through! I think that your suggestion makes sense, but we'll wait for @jpanther's feedback as he's the author and maintainer.

jpanther:
Thanks for this suggestion and subsequent discussion, @OCram85 and @fteem. Apologies for my delayed reply as I've been inundated with work the past couple of weeks!

I do like the idea of adding something along the lines of this change although I wonder if it would be better to bundle all the functionality into the existing `alert` shortcode? For instance, `alert` could take an additional parameter for the icon to be used.

```go
{{< alert "note" >}}
This is a note.
{{ /alert }}
```

As you've pointed out, adding the prefix "NOTE:" into the shortcode is also problematic as it requires localisation. I understand what you mean when you say you prefer to include as much in the shortcode as possible but isn't that text best included as part of the content?

It is far more flexible to use the syntax below - it keeps all the content together in the markdown file and it allows portability if you want to move to another theme and keep the context with the note.

```go
{{< alert "note" >}}
**NOTE:** This is a note.
{{ /alert }}
```

I'm not sure what a lot of other themes do in this regard, but I have seen [`admonition`](https://hugoloveit.com/theme-documentation-extended-shortcodes/#4-admonition) from the LoveIt theme which provides something similar to what we're describing here.

Ultimately my goal with any change like this is to keep it simple for the end user so one shortcode that can handle multiple use cases is ideal.

OCram85:
So what's about this:

- If the `alert` shortcode would be renamed into something more generic like _card_, _box_, _banner_ ...
- We could add an property called `severity` = `(info | note | alert | ....)` to mimic different styles.
- This could include the icon an background color to get unique look and feel.

As already discussioned here, there could be 2 possible ways to add an prefix per severity:

1.  Just let the user write the prefix like:  **Note:** , **Alert:** , **Info:**
2. derive the prefix from the used severity type and integrate it with i18n values into the shortcode.

fteem:
Humbly, I think that we should just copy what admonition does and be done with it. But whatever @jpanther decides, I am sure it will look and work great. Looking forward to it!

jpanther:
Okay, I think I lead you astray by mentioning admonition. I don't want to copy that style or functionality at all because it is not in fitting with the design of this theme. I have no desire to introduce a multitude of colours like that and I think we already have enough icons in the theme to handle a wide variety of use cases (although I'm happy to add some more generic ones).

I feel like this is the sort of addition where if this level of complexity is something you really need, it's probably something that should be added locally at the project level. I do this already in my own Congo project where I have a custom version of the alert shortcode (you can [see it in action here](https://jamespanther.com/writings/i-switched-from-google-analytics-to-fathom-analytics/)). I purposefully didn't include this in the core project because I don't think it's generic enough to fit with a wide variety of use cases.

I still think it's worth adding extra functionality to the existing shortcode but I'm not convinced we're on the right track with all these suggestions. Perhaps we should go back to the start and determine what is the actual problem this change is trying to solve? What is the use case for all these different boxes within the content?

If I were to propose this sort of architecture, is this something that would meet 95% of those use cases?
- add optional parameter for icon
- add optional parameter for background (which can either be the neutral, primary or secondary colour that is already in the theme)

By adding just two parameters like this the box gains a significant boost to the number of variations available, while still being simple enough that most users can use it without these parameters and achieve utility.

You could then do something like the below to get two very different looking boxes:

```go
{{< alert >}}
**IMPORTANT:** This message stands out.
{{< /alert >}}

{{< alert icon="note" variant="neutral" >}}
**NOTE:** This is a subdued looking message.
{{< /alert >}}
```

OCram85:
If it's ok for you, then of course it works for us, too 😀 
I've only tried to find a way to simply add different alert variants - Didn't expect to start such a complex discussion ^^

As you already explained, this should work perfect for us:

```
{{< alert icon="note" variant="neutral" >}}
**NOTE:** This is a subdued looking message.
{{< /alert >}}
```

OCram85:
I guess this PR can be closed

PR #144, opened 2022-03-07
Adds pt-PT translation

brunomiguel:
Adds the European Portuguese translation

The theme already has the Brazilian Portuguese. This commit adds the European Portuguese translation


jpanther:
Thanks for taking the time to translate the theme, @brunomiguel! I'll merge these changes in.

PR #145, opened 2022-03-07
Use the group by year var in the term partial

igalarzab:
hey!

Thanks for open-sourcing this theme :)

This PR just adds the `groupByYear` var in the `term` layout so we could disable the grouping in the taxonomy pages


jpanther:
Thanks for fixing this, @igalarzab! I'll merge this change in and it will be included the next release.

PR #155, opened 2022-03-22
💄 Hide navigation during printing

MikiVanousek:
Printing an article will now result in a clean, text only document.


jpanther:
Thanks! This is a good improvement. I'll merge it in now and it will be part of the next release.

PR #163, opened 2022-03-31
:globe_with_meridians: Add hebrew i18n

MoshiBin:
This change adds a `he.yaml` file for Hebrew language support. It also updates the `configuration` page and adds Hebrew to the list of supported languages.


jpanther:
Thanks for adding these translations and for updating the relevant docs. I'll merge this in and it will be available in the next release.

PR #165, opened 2022-04-01
🌐 Fix footer menu display in RTL

MoshiBin:
The footer menu did not have these `rtl:` modifiers to change the margin direction. This PR adds the modifiers for both LTR and RTL, similar to how it's done in `header.html`.


jpanther:
Thanks for picking up on this one. I'll merge this fix in.

PR #187, opened 2022-05-01
👪 Update users.md

max-alletsee:


jpanther:
Thanks for using Congo! It's really interesting to see it being used for a food blog! I'll merge this in and the link will appear on the public site when the next theme update is released.

PR #188, opened 2022-05-03
docs: add www.guyu.me to user.md

wizardbyron:


jpanther:
Thanks for using Congo. I had a quick look at your link and it seems the URL you've provided isn't working. Can you please check to make sure it's correct?

younsl:
It needs to change guyu.me to www.guyu.me.
www.guyu.me works for me.

PR #190, opened 2022-05-09
📝 Add user to users.md

younsl:
* Add congo theme user


jpanther:
Thanks for using Congo, @seyslee! I'll merge this change in and it will go live when the next theme release is published.

PR #191, opened 2022-05-09
:memo: Update configuration document

younsl:
#### Why we need it
Some users may want a simpler footer without theme information message.

#### Changes
* Add `attribution` value to configuraiton doc


jpanther:
Thanks for suggesting this change, however this is not intended to be a feature as it's something that is not fully implemented into the theme. The parameter has been undocumented since the beginning as it was simply something that I left in the code without fully exploring it. However, as people have slowly discovered its existence, it has begun raising more and more questions.

Although I don't currently intend to keep the parameter around, I do think that something needs to take its place. I've [started a discussion](https://github.com/jpanther/congo/discussions/192) to ask the community for feedback and I'd encourage you to contribute your thoughts there.

jpanther:
I've added some new parameters to the dev branch in 31c7cca31dc27b0dbb86cbbb891a48230e18a1e3 that will allow better customisation of the site footer. As part of this I have removed the `attribution` parameter and replaced it with a more appropriately named `showThemeAttribution` parameter, along with other style and supporting changes to improve how this functionality works. These new parameters are fully documented and supported.

In light of this change, I'm going to close this PR as it's no longer relevant.

younsl:
LGTM, thanks @jpanther!

PR #197, opened 2022-05-14
add datanalyze.be to users.md

AntoineSoetewey:


jpanther:
Thanks, @AntoineSoetewey! I'll merge this in. The public page will be updated when the next theme release is published.

PR #198, opened 2022-05-15
Update example baseUrl in getting started

bortox:
Just in case someone needs it, writing https://your_domain.com as baseURL makes the function fetchJSON in `assets/js/search.js` fetch an url like https://your_domain.comindex.json, so in this scenario the search function does not load the index file since it is not found.

I had the search function not working on my website and later found out the cause, it's just the slash that I forgot to add to the baseURL.


jpanther:
Thanks for picking up on this. I'll merge the change in.

PR #221, opened 2022-06-19
fix Netlify preview deployments

jkpe:
Netlify uses an old version of Hugo (0.8x) unless you tell it to use something newer. 0.8x does not support congo as a module so you'll end up with this error in your preview deployments.

```
WARN 2022/06/19 17:47:35 Module "github.com/jpanther/congo/v2" is not compatible with this Hugo version
```


jpanther:
Thanks for this addition. I'm happy to merge this in. Note that the public documentation will not update to reflect this change until the next theme release is published.

Issue #223, opened 2022-06-21
Search not working when deployed to Netlify


eliaserland:
### What happened?

# Problem
The search bar on my website functions normally when served locally, but does not work when deployed to Netlify.  When the search icon is clicked, the search bar appears seemingly as per normal, but it doesn't do anything, no matter what is typed into the search box.

# Links
Website: [https://utvik.netlify.app](https://utvik.netlify.app)
Repo: [https://github.com/eliaserland/utviksbygdegard](https://github.com/eliaserland/utviksbygdegard)

# Investigation
If I do an F12 in my browser I get this error message when I click the search icon:
`Cross-Origin Request Blocked: The Same Origin Policy disallows reading the remote resource at https://utvik.netlify.appindex.json/. (Reason: CORS request did not succeed). Status code: (null).`

I do not entirely understand the meaning of this error message, but what I can notice is that there is a missing forward slash between the website URL `utvik.netlify.app` and the filename `index.json`, which likely seems to break this URL request.

If I trace this back to the congo repo, the only reference to `index.json` I can find is appearing in file `assets/js/search.js` on line 110 where I believe the search index is built:
```
function buildIndex() {
  var baseURL = wrapper.getAttribute("data-url");
  fetchJSON(baseURL + "index.json", function (data) {
    var options = {
      shouldSort: true,
      ignoreLocation: true,
      threshold: 0.0,
      includeMatches: true,
      keys: [
        { name: "title", weight: 0.8 },
        { name: "section", weight: 0.2 },
        { name: "summary", weight: 0.6 },
        { name: "content", weight: 0.4 },
      ],
    };
    fuse = new Fuse(data, options);
    indexed = true;
  });
}
```
There seems to be a concatenation of strings between `baseURL` and `"index.json"` happening, which is where I believe I would need a forward slash inserted in order for the URL request not to break.

I'm not certain that this will fix my issue of search not working, but it should however likely be a step in the right direction.

Since I don't know JavaScript, I don't feel comfortable enough to open a pull request where I have inserted a single `/` in front of the `index.json` string. Spontaneously, I feel like the rigorous solution should be something like verifying that the variable `baseURL` contains a single trailing forward slash, but I don't know enough JavaScript to implement this myself in a robust way.

Btw, thanks for a great theme with amazing documentation, this deserves major props.

### Theme version

v2.2.2

### Hugo version

v0.100.2 on Netlify, v0.96.0 locally

### What browsers are you seeing the problem on?

Firefox

[...]

PR #224, opened 2022-06-21
:bug: Fix issue #223, ensure trailing forward slash baseURL during search index build

eliaserland:
Fix to issue #223.

Should now hopefully verify that `baseURL` contains a trailing forward slash. This fixes my personal issue with search not working, hopefully it doesn't break anyone else's search.


jpanther:
Thanks for this fix, @eliaserland. I've tested the change and it works. I can't see any reason why this would break any other build configurations as the base URL should always end in a forward slash so I'll go ahead and merge this in.

PR #234, opened 2022-06-29
feat: :+1:  add Japanese language :jp:

shin-sforzando:
Dear @jpanther ,

I have added the Japanese language file with reference to the English language file.
However, the words in the English language file are familiar to Japanese people, so I don't feel it is necessary to dare to convert them to Japanese.
If someone uses this Japanese language file, `dateFormat` in `languages.ja.toml` should be like `"2006年01月02日"` to be closer to the Japanese convention.
Anyway, I hope it helps you.

P.S.

Screen Shot 2022-06-29 at 13 00 10

```ja.yaml
footer:
  dark_appearance: "ダークモードへ切り替え"
  light_appearance: "ライトモードへ切り替え"
```

Unfortunately, I was unable to complete the verification for the footer appearance switching section.
It keeps showing "Switching to dark" even when switching from dark to light.
Is this happening only in my environment?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ## [Unreleased]
 
+### Added
+
+- Japanese translation ([#234](https://github.com/jpanther/congo/pull/234))
+
 ### Fixed
 
 - Main content misaligned when hamburger menu is opened at large viewport sizes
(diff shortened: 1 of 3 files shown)


jpanther:
Thanks for the Japanese translation @shin-sforzando! I have often wondered if I should move the `dateFormat` out of the site config and into the i18n file but I figure most people like to have control over how dates are displayed. I'll merge these changes in and they will be included in the next release.

That issue you are having with the appearance switcher is not just you. It looks like you've found a bug!

shin-sforzando:
I always appreciate your great themes.
And I love Factorio, too:)

PR #244, opened 2022-07-10
Add KaTeX inline delimiters $$

ruihao-li:
Currently, the only option to render inline math equations with KaTex is the delimiters `\\(` and `\\)`. This is unnatural to LaTeX users and quite cumbersome to write in math-heavy documents. Therefore, I propose to add the more nature delimiters `$` and `$` for inline equations, e.g., `$E = mc^2$`.


jpanther:
Thanks for the suggestion, however this change is modifying a bundled vendor library directly, which is not acceptable. In order to add support for additional delimiters they should be passed in to the auto-renderer at page load as per https://katex.org/docs/autorender.html. This change therefore requires a new javascript file to be added into the theme JS assets bundle at build time in order to provide the necessary parameters. The challenge here is that the bundle is built based upon your site configuration but the Katex library is only included on a per-page basis.

Although I am not a heavy Katex user, it does seem strange to me that the `$` delimiters are 'more natural' yet they are not included in the library defaults. As there are potential performance trade-offs, perhaps more discussion is required to see if others would find this change beneficial?

A simpler solution for changing this behaviour is to override it yourself in the `katex.html` partial. A local partial in your project with the following should be enough to get your desired outcome:

```html
<script>
    document.addEventListener("DOMContentLoaded", function() {
        renderMathInElement(document.body, {
          delimiters: [
              {left: '$$', right: '$$', display: true},
              {left: '$', right: '$', display: false},
              {left: '\\(', right: '\\)', display: false},
              {left: '\\[', right: '\\]', display: true}
          ],
          throwOnError : false
        });
    });
</script>
```

ruihao-li:
@jpanther Thanks for the detailed explanation and suggestions! I'm new to the web development sort of things, so sorry about the rookie mistake here. 

I guess `$..$` is not "more natural" than `\(...\)` but just more convenient. TeX supports both delimiters and I just happen to have been using `$...$` for years, which makes it look "more natural" to me. MathJax explains why it does not define the inline delimiter `$...$` by default [here](https://docs.mathjax.org/en/latest/input/tex/delimiters.html). Maybe KaTeX does it for the same reason. I agree that it would be better to hear what others think about adding this option.

jpanther:
Thanks for the link to that explanation. I think given the fact this change could cause unintended consequences for anyone inserting a `$`, I'm going to close this PR. It's probably best for individual users to adjust the defaults in their own projects as per my suggestion above if they wish to override this behaviour.

PR #251, opened 2022-07-28
Please add my site to the list in users.md

wizardbyron:


jpanther:
I've checked the deploy preview and the submitted URL doesn't appear to work. Could you please amend the PR with the correct URL?

younsl:
This PR is duplicated by #188.
As I checked, You need to change the requested URL to https://www.guyu.me for it to work properly.

PR #252, opened 2022-07-28
Please add my website to your list

hotlittlewhitedog:


jpanther:
Thanks for using Congo! I'll merge this change in and the public list will update when the next theme release is published.

PR #262, opened 2022-08-04
🌐 Add Traditional Chinese Translation

tomy0000000:
Adding the translation of Traditional Chinese, a different variant of Chinese


jpanther:
Thanks for the new translation, @tomy0000000! I'll merge this change in and it will be part of the next release.

PR #268, opened 2022-08-08
feat: :+1: introduce prettier-plugin-tailwindcss

shin-sforzando:
Hello @jpanther

Congo is using Prettier and Tailwind CSS.
Then you could standardize the coding style of Congo users by installing [tailwindlabs/prettier-plugin-tailwindcss](https://github.com/tailwindlabs/prettier-plugin-tailwindcss).
I used it for about a month in the development of my own site using Congo and had no problems in my environment.
I think it is great that it is an official plugin provided by Tailwind and that we don't have to configure anything just by installing it.
I would be glad if you would consider introducing this plugin.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "hugo-congo-theme",
-  "version": "2.3.0",
+  "version": "2.3.1",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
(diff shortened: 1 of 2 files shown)


jpanther:
Thanks for the suggestion. I use the official Tailwind extension for VSCode which I believe applies the same coding standards so I can't see any reason to not add this package to allow others to benefit from it. I'll merge this in.

PR #277, opened 2022-08-17
Fix typo in docs

max-alletsee:


jpanther:
Thanks for fixing this, @max-alletsee! I'll merge this change in.

PR #281, opened 2022-08-23
Added my personal site to users.md

nunocoracao:
Thanks for the hard work, theme looks awesome. Did a couple of customization on my own might suggest some of them evetually...


jpanther:
Thanks for using Congo, @nunocoracao! I really like what you've done with your customisations. I'll merge this change in and your site will appear on the public list when the next theme release is published.

PR #282, opened 2022-08-23
Add a new header type: hybrid so that the UI can switch between bas…

nunocoracao:
…ic and hamburger depending on device screen size.

Created new partial to support use-case
Updated CSS to include the hide classes to be used depending on the screen size
Updated the search.js code to support two different buttons at the same time in the UI

### DIFF
--- a/assets/css/compiled/main.css
+++ b/assets/css/compiled/main.css
@@ -1845,6 +1845,28 @@ body:has(#menu-controller:checked) {
           text-decoration-line: underline;
 }
 
+/* Responsive hides for hybrid header */
+
+@media screen and (min-width: 0px) {
+  .hidebasic {
+    display: none !important;
+  }
+
+  .hidehamburger {
+    display: flex !important;
+  }
+}
+
+@media screen and (min-width: 1250px) {
+  .hidebasic {
+    display: flex !important;
+  }
+
+  .hidehamburger {
+    display: none !important;
+  }
+}
+
 .pointer-events-none {
   pointer-events: none;
 }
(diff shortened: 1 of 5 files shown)


jpanther:
Thanks for the PR, @nunocoracao. I really like this feature however I'm not sure why you chose to implement the CSS yourself rather than using the Tailwind CSS classes? I prefer to always use the Tailwind classes when they are available and I think there are existing classes for all the changes you have proposed.

I also wonder if there is a way to implement this without so much code duplication in the HTML? For instance, is it possible to show/hide the components at different viewport sizes while still maintaining the same DOM structure?

nunocoracao:
Hey @jpanther, thanks for your comments. I think both of your suggestions are awesome. I think the main reason is that I am not that proficient with tailwind so this is just the way I "hacked" it for my personal website. Just thought or sharing. I can try to improve your poitns but if you could help it would be amazing.

nunocoracao:
@jpanther tried the css classes in the .html and was not able to get it to work (e.g. md:visible and what not). Ended up replacing the css by tailwind commands which is a little bit better.

jpanther:
I'll keep this one open for now. I do want to work on it but haven't had time.

jpanther:
Thanks for your patience with this request, @nunocoracao. I've finally had some time to get a hybrid layout working and it's now in the `dev` branch. As it's a different implementation to the one proposed in this PR, I am going to close this without merging, however I expect that the new layout will meet your use case. If you have any feedback, please let me know!

PR #288, opened 2022-09-02
add hungarian sports association using this theme

nandor-magyar:
First time experience with hugo. This theme is slick and awesome, it was a breeze, it is brilliant really. Thank you for making it.
Keep up the good work!
All the best,
n.


jpanther:
Thanks for using Congo, @nandor-magyar! I'll merge this change in and your site will appear on the public list when the next theme release is published.

PR #289, opened 2022-09-02
Fix error calling Paginate

michaeldorner:
Fixes issues of hugo v0.102 throwing the following error: `_default/term.html:16:14: executing "main" at <.Paginate>: error calling Paginate: cannot convert type page.PagesGroup to Pages`

I think it was caused by no `terms` set up, but I am no Hugo expert. Could please someone check if I fixed the issue properly?

### DIFF
--- a/layouts/_default/term.html
+++ b/layouts/_default/term.html
@@ -13,15 +13,17 @@
     </section>
   {{ end }}
   <section>
-    {{ range (.Paginate (.Pages.GroupByDate "2006")).PageGroups }}
-      {{ if $.Params.groupByYear | default ($.Site.Params.list.groupByYear | default true) }}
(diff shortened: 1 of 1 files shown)


jpanther:
Thanks for the PR, @michaeldorner. I've been trying to replicate the issue you are referring to but haven't been able to get that error to appear. Are you able to describe the configuration you're using when it appears? I'm using Hugo v0.102.3 in my testing.

michaeldorner:
Thanks for looking at the issue. I started with a plain, empty website. Then the problem described above occurred. I can share the current state of my website if this helps.

jpanther:
Thanks @michaeldorner, I've merged this fix in.

PR #291, opened 2022-09-05
Update users.md

ghost:
Add jcransom.com as personal site and blog


jpanther:
Thanks for using Congo, @jcransom! I'll merge this change in and your site will appear on the public list when the next theme release is published.

PR #310, opened 2022-10-01
:memo: Add my website to users.md

MikiVanousek:


jpanther:
Hey @MikiVanousek, thanks for using Congo. While I'm happy to add your site to the list, I do think that the description you've chosen for your entry is in poor taste. Is that meant to be funny?

MikiVanousek:
Yeah, it was supposed to be a joke. I understand if you don't like it. I will come up with something better.

PR #317, opened 2022-10-13
🩹Fix #313 ordering content in a list layout

### LINKED ISSUE #313: Ordering content in a list layout
### What happened?

Hey! The theme is awesome. I faced some minor issues ordering content in a list layout. I'm not sure if they can be considered bugs but I wanted to share.

#### Order by date before weight

While default Hugo behavior is to order content by [weight > date > title > filename](https://gohugo.io/templates/lists/#default-weight--date--linktitle--filepath) Congo seems to order content by date > weight > ... So if `date` exists in a frontmatter `weight` is ignored.

#### Order by weight goes bottom-to-top

When there is no dates in posts Hugo orders content by weight. But content with lower weight comes lower in the list while Hugo docs says ["Lower weight gets higher precedence."](https://gohugo.io/templates/lists/#by-weight). So I expected a page with the lowest weight to appear first but it appears last in a list with the Congo theme.

### Theme version

v2.3.0

### Hugo version

v0.101.0

Chrome

### Relevant Hugo log output

_No response_

gtitov:
Fix #313. PR sets pagination and ordering to Hugo's default if `groupByYear` parameter is set `false`. It makes the Congo theme follow [Hugo's convention of ordering content inside list templates](https://gohugo.io/templates/lists/#order-content).

### DIFF
--- a/layouts/_default/list.html
+++ b/layouts/_default/list.html
@@ -26,14 +26,18 @@
   </section>
   {{ if gt .Pages 0 }}
     <section>
-      {{ range (.Paginate (.Pages.GroupByDate "2006")).PageGroups }}
-        {{ if $.Params.groupByYear | default ($.Site.Params.list.groupByYear | default true) }}
+      {{ if $.Params.groupByYear | default ($.Site.Params.list.groupByYear | default true) }}
+        {{ range (.Paginate (.Pages.GroupByDate "2006")).PageGroups }}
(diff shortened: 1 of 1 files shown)


jpanther:
Thanks for reporting this and supplying a fix, @gtitov. I'm happy with this change and will merge it in.

gtitov:
And I'm happy to contribute to the wonderful product!

PR #322, opened 2022-10-19
✨ Move logo to its own partial

mhulse:
This allows for the logo to be overridden more easily without having to replace all of `hamburger.html` and `basic.html`.


jpanther:
Thanks for the PR. I've made a minor change and moved the logo file to the root partials directory so that it's not accidentally called as a header in a misconfigured site. Other than that, I'm happy to accept this change and merge it in.

PR #326, opened 2022-10-20
:sparkles: Independent control of taxonomy display on list AND single pages

mhulse:
**Sensible defaults:** `.Params.showTaxonomies` will trump `.Site.Params.article` and `.Site.Params.list` `.showTaxonomies` settings.


jpanther:
Thanks for the PR, @mhulse. I think this change adds useful functionality so I'll accept it. Before I merge I'm just going to reorder the parameters so that they match the existing article ones and include some documentation.

mhulse:
Thanks!

PR #332, opened 2022-10-23
:memo: Add major.io to users list

major:
I love using Congo! 💙


jpanther:
Thanks for using Congo, @major! I'll merge this change in and your site will appear on the public list when the next theme release is published.

major:
Thanks, @jpanther! 🫂

PR #362, opened 2022-11-10
Added cgutierr-zgz.github.io to users

cgutierr-zgz:
Added my my github.io to users.md


jpanther:
Thanks for using Congo, @cgutierr-zgz! I'll merge this change in and your site will appear on the public list when the next theme release is published.

PR #378, opened 2022-11-21
Fix: unique meta descriptions from articles

### LINKED ISSUE #377: `meta name=description` not set from article description (or summary)
### What happened?

Hi!

I noticed the lighthouse SEO scores aren't 100% if you don't configure a (Site-wide).

> Document does not have a meta description
> Meta descriptions may be included in search results to concisely summarize page content. [Learn more](https://web.dev/meta-description/?utm_source=lighthouse&utm_medium=lr).

```toml
# config/_default/languages.en.toml
title = "My Blog"
# description = "blah"
```

Uncommenting the above appears to set the `<meta name="description" content="blah">` to be the same for all pages. This is not advisable, according to [this article](https://www.searchenginejournal.com/matt-cutts-answers-every-page-needs-unique-meta-description/78133/)

For the lazy, like me, inserting the  [.Page.Summary](https://gohugo.io/variables/page/). Which it appears is what the internal [OpenGraph](https://github.com/gohugoio/hugo/blob/master/tpl/tplimpl/embedded/templates/opengraph.html) template does. In fact, you can see this already working on each article page with `<meta name="og:description" content="the summary">`

Inspecting the variables that the [internal template for og:description](https://github.com/gohugoio/hugo/blob/master/tpl/tplimpl/embedded/templates/opengraph.html), I see

```html
# layouts/partials/extend-head.html
<meta name="description"
  content="{{ with .Description }}{{ . }}{{ else }}{{if .IsPage}}{{ .Summary }}{{ else }}{{ with .Site.Params.description }}{{ . }}{{ end }}{{ end }}{{ end }}" />
```

This defaults to the 
  1. optional frontmatter `description:` or
  2. the (autogenerated) Page Summary, or 
  3. the site config (assumes duplicate meta tags are better than nothing?)

But for meta tags to work the same way in Congo, I believe we would need this (breaking?) change:

```diff
  # layouts/partials/head.html
  {{ if templates.Exists "partials/extend-head.html" }}
-  {{ partialCached "extend-head.html" .Site }}
+  {{ partial "extend-head.html" . }}
  {{ end }}
```

I'm happy to send in an MR,  but wanted to discuss the preferred fix first.

### Theme version

2.0.5

### Hugo version

0.104.3

Firefox

### Relevant Hugo log output

_No response_

glennr:
fixes :bug: #377

The Meta description tag is now generated from either
1. page description, or
2. Page content summary, or
3. Site wide config

Before

image

After

image

### DIFF
--- a/layouts/partials/head.html
+++ b/layouts/partials/head.html
@@ -14,9 +14,7 @@
     <meta name="title" content="{{ .Title | emojify }} &middot; {{ .Site.Title | emojify }}" />
   {{- end }}
   {{/* Metadata */}}
-  {{ with .Site.Params.description -}}
-    <meta name="description" content="{{ . }}" />
-  {{- end }}
+  <meta name="description" content="{{ with .Description }}{{ . }}{{ else }}{{if .IsPage}}{{ .Summary }}{{ else }}{{ with .Site.Params.description }}{{ . }}{{ end }}{{ end }}{{ end }}" />
   {{ with .Site.Params.keywords -}}
     <meta name="keywords" content="{{ . }}" />
   {{- end }}
--- a/layouts/partials/head.html
+++ b/layouts/partials/head.html
@@ -14,7 +14,7 @@
     <meta name="title" content="{{ .Title | emojify }} &middot; {{ .Site.Title | emojify }}" />
   {{- end }}
   {{/* Metadata */}}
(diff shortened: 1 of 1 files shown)


jpanther:
Thanks, @glennr! I'm happy to accept this change and will merge it in now.

PR #397, opened 2022-12-02
:memo: Add dat-adi to the users

dat-adi:
Added my site to the `users.md`, and fixed an alignment. Thank you for providing this masterpiece! :D


jpanther:
Thanks for using Congo, @dat-adi! I'll merge this change in and your site will appear on the public list when the next theme release is published.

PR #398, opened 2022-12-03
🌐 Add Indonesian translation (id.yaml) & 📝 shim to the users

ghost:


jpanther:
Thanks for the translation, @5h1m! I'm happy to accept and merge this change in.

PR #427, opened 2022-12-28
:bug: Fix render-image.html for SVG

### LINKED ISSUE #413: Svg images in page bundle
### What happened?

Including an SVG from a page bundle results in an error due to use of Resize method on none raster image. The image can be included fine from static.

### Theme version

v2.4.2

### Hugo version

v0.108.0

_No response_

### Relevant Hugo log output

```shell
"/tmp/hugo_cache/modules/filecache/modules/pkg/mod/github.com/jpanther/congo/v2@v2.4.2/layouts/_default/_markup/render-image.html:21:14": execute of template failed at <.Resize>: error calling Resize: this method is only available for raster images. To determine if an image is SVG, you can do {{ if eq .MediaType.SubType "svg" }}{{ end }} 

/tmp/hugo_cache/modules/filecache/modules/pkg/mod/github.com/jpanther/congo/v2@v2.4.2/layouts/_default/_markup/render-image.html:21:14:

        class="my-0 rounded-md"
        srcset="
          {{ (.Resize "330x").RelPermalink }} 330w,
          {{ (.Resize "660x").RelPermalink }} 660w,
          {{ (.Resize "1024x").RelPermalink }} 1024w,
```

kpavlov:
SVG images do not support resizing, so `srcset/src` attributes should be simplified

Fixes #413

### DIFF
--- a/layouts/_default/_markup/render-image.html
+++ b/layouts/_default/_markup/render-image.html
@@ -15,16 +15,24 @@
   {{ end }}
   {{ with $resource }}
     <figure>
-      <img
-        class="my-0 rounded-md"
-        srcset="
-          {{ (.Resize "330x").RelPermalink }} 330w,
-          {{ (.Resize "660x").RelPermalink }} 660w,
-          {{ (.Resize "1024x").RelPermalink }} 1024w,
-          {{ (.Resize "1320x").RelPermalink }} 2x"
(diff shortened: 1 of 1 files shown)


jpanther:
Thanks for the PR, @kpavlov! I'm happy to accept this fix and will merge it in now.

PR #429, opened 2022-12-31
Add my personal site to sample users

FlorianPfisterer:


jpanther:
Thanks for using Congo, @FlorianPfisterer! I'll merge this change in and your site will appear on the public list when the next theme release is published.

PR #435, opened 2023-01-09
🌐 Add new sharing.mastodon i18n string

ghost:


jpanther:
Thanks for the quick update, will merge this in now.

PR #436, opened 2023-01-10
Add davidrothera.me to users.md

davidrothera:
Adding new site to the users list

👍 thanks for the awesome work @jpanther


jpanther:
Thanks for using Congo, @davidrothera! I'll merge this change in and your site will appear on the public list when the next theme release is published.

PR #438, opened 2023-01-11
Add aria-label to appearance switcher button

davidrothera:
At the moment we get knocked on the a11y portion of a lighthouse report if using this feature (screenshot below)

image

This PR adds an `aria-label` which resolves the issue.


jpanther:
Thanks, @davidrothera. I'm always happy to accept changes that improve accessibility. I'll merge this in.

PR #439, opened 2023-01-11
Fix a11y issues with article-link

davidrothera:
At the moment there are two lighthouse issues:
* Image (if present) doesn't have an alt tag
  * Fixed by adding a blank `alt` tag as the image is purely decorative as is described in standards
* Link has no name
  * Use the same name as the title in an `aria-label`

Screenshot 2023-01-11 at 23 33 37


jpanther:
Thanks again, @davidrothera! I'll merge this in now.

PR #440, opened 2023-01-14
Added ethantroy.com

ethanolivertroy:
.


jpanther:
Thanks for using Congo, @ethanolivertroy! I'll merge this change in and your site will appear on the public list when the next theme release is published.

PR #446, opened 2023-01-17
🛠 Links render hook should add `noreferrer noopener`

tomy0000000:
According to [this](https://www.jitbit.com/alexblog/256-targetblank---the-most-underestimated-vulnerability-ever/) and [this](https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/jsx-no-target-blank.md), it is best if the links with `target="_blank"` also to have `rel="noreferrer noopener"` set to prevent any security issue.


jpanther:
Thanks for the change, @tomy0000000. I'm happy to accept and include this in the next release. I'll merge it in now.

ghost:
Hi, I know this is old but the additional rel="noreferrer noopener" does not work in menus.  Showing on mastodon link in th menu on mark.hysted.com

PR #448, opened 2023-01-18
🌐 Update `zh-TW` translation

tomy0000000:
Add Traditional Chinese translation for `article.reading_time.one` and `sharing.mastodon`


jpanther:
Thanks for these updates, @tomy0000000. I'll merge them in now.

PR #450, opened 2023-01-19
Update index.md

maijau:
Added personal blog to user list.


jpanther:
Thanks for using Congo, @maijau! I'll merge this change in and your site will appear on the public list when the next theme release is published.

PR #471, opened 2023-02-12
correct typo and add site

AntoineSoetewey:


jpanther:
Thanks, @AntoineSoetewey! I'll merge this change in and they will appear in the public docs when the next theme release is published.

PR #502, opened 2023-03-16
Fixes for Russian (ru) translation

ementat:
Some fixes for Russian (ru) translation.


jpanther:
Thanks for these fixes, @ementat. I'll merge them in now.

PR #508, opened 2023-03-23
🐛 Fix broken `srcset` links when using hugo with `--minify` option

uschmidt83:
HTML code generated by templates `layouts/_default/single.html` and `layouts/partials/article-link.html` caused missing space between the image link and the associated intrinsic width (e.g. `...box.jpg1024w` instead of `...box.jpg 1024w`) if hugo was used with the `--minify` option.

Also updated `layouts/_default/_markup/render-image.html` and `layouts/shortcodes/figure.html` to not cause line breaks in the srcset markup.

### DIFF
--- a/layouts/_default/_markup/render-image.html
+++ b/layouts/_default/_markup/render-image.html
@@ -24,10 +24,10 @@
             src="{{ .RelPermalink }}"
           {{ else }}
             srcset="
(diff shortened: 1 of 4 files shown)


jpanther:
Thanks for this fix, @uschmidt83! I'm happy to accept this change and will merge it in now.

PR #511, opened 2023-03-24
✨ Respect `showDate` in search results

uschmidt83:
Also hide date for list pages, which don't have their date shown in the theme.


jpanther:
Thanks for this change, @uschmidt83. I think it makes sense to hide these dates on the search results when they aren't visible elsewhere so I'm happy to merge this in.

PR #527, opened 2023-04-06
author.image SmartCrop Issue

### LINKED ISSUE #526: Author Image file cropped
### What happened?

When Using `author.image` the image seems to be cropped, then rounded. It is not css as the image provided is already cropped. 
Everything done accoring to instructions
```
/assets/ 
| ---- img/  
       | -- problem.jpg  
```
`  image = "img/problem.jpg"`

Input image:
problem
Page:
image
Element:
image
Hugo shows no errors 

### Theme version

v2.5.3

### Hugo version

v0.111.3+extended windows/amd64

Firefox, Chrome, Microsoft Edge

### Relevant Hugo log output

_No response_

mchlebosz:
When using author.image the image seems to be cropped, then rounded. It is not CSS as the image provided is already cropped.
The problem seems to be the SmartCrop feature of Hugo #526 

The solution here is to define crop box placement explicitly: `$authorImage.Fill "192x192 Center"` 
That way, no part of an image is lost when a 1:1 image is provided.

### DIFF
--- a/layouts/partials/author.html
+++ b/layouts/partials/author.html
@@ -3,7 +3,7 @@
     {{ with .Site.Author.image }}
       {{ $authorImage := resources.Get . }}
       {{ if $authorImage }}
-        {{ $authorImage := $authorImage.Fill "192x192" }}
+        {{ $authorImage := $authorImage.Fill "192x192 Center" }}
         <img
           class="!mb-0 !mt-0 me-4 h-24 w-24 rounded-full"
           width="96"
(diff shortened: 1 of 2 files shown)


metbril:
Keep?

jpanther:
Thanks for this fix, @mchlebosz. Apologies for taking so long to review the PR, I've been super busy! I think this should probably be a configurable parameter in the future, but for now this is a sensible default so I'm happy to merge it in.

PR #528, opened 2023-04-07
Add andrew-jones.com to user list

andrewrjones:
Have been using for nearly a year. Thanks for a great theme!


jpanther:
Thanks for using Congo, @andrewrjones! I'll merge this change in and your site will appear on the public list when the next theme release is published.

PR #540, opened 2023-04-24
Add nikita.computer to the user sites page 

nikitawootten:
I've been using this theme for a while now. Thanks for the amazing theme.


jpanther:
Thanks for using Congo, @nikitawootten! I'll merge this change in and your site will appear on the public list when the next theme release is published.

PR #541, opened 2023-04-25
add translation for slovak [sk] and czech [cs] language

matusstafura:


jpanther:
Thanks for these translations, @matusstafura! I'll merge them in and they'll be packaged into the next release.

PR #544, opened 2023-04-28
Update Dutch (nl) translation 🇳🇱

metbril:
Translated new Mastodon string.


jpanther:
Thanks for the updated string, @metbril. I'll merge it in.

PR #552, opened 2023-05-08
✨ feat: Lazy load image by default

tomy0000000:
Make lazy loading images by default to reduce the load time on pages with many pictures.


jpanther:
Thanks for the PR. I do wonder if this is something that should be added as a configurable parameter? Also, is there a reason you chose to only apply lazy loading to the markdown rendering template and not in other areas like the `figure` partial?

tomy0000000:
> I do wonder if this is something that should be added as a configurable parameter?

Sounds reasonable, should probably work on this.

> is there a reason you chose to only apply lazy loading to the markdown rendering template and not in other areas like the figure partial?

Not really, that's my mistake in missing those, which should be set to lazy-load, too.

Seems like we need to work on this more to support, I'll try to find some time.

tomy0000000:
@jpanther I think this is ready to go!

jpanther:
Thanks for the changes, I think this is now a complete implementation and am happy to merge it in. It will be included in the next theme release.

ghost:
Welcome to getting moaned at by GTMetrix (Don't lazy load Largest Contentful Paint image) and Pagespeed (Largest Contentful Paint image was lazily loaded) for lazy loading.  Above the fold shouldn't be lazy loaded, it actually just looks weird , and shifts text down when it appears.

Better would be to not lazy load cover or featured image, then load rest - or make it option. none/all/notcover?

jpanther:
Thanks @hystedm. I've opened a new issue #591 to track this.

PR #558, opened 2023-05-23
Add blog.dejavu.moe to the user sites page

DejavuMoe:
I'm using the hugo congo theme, thanks for your great work, thanks!


jpanther:
Thanks for using Congo, @DejavuMoe! I'll merge this change in and your site will appear on the public list when the next theme release is published.

PR #568, opened 2023-06-01
🛠 fix: icon padding in menus

tomy0000000:
The menu items with icons currently have a small spacing between the icon and the text, which is nice. However, the spacing exists because the rendered HTML has much spacing between the `<span>`s element, and this spacing will disappear when passing the `--minify` flag to `hugo serve` or `hugo build` commands, causing the icon to stick close to the text.

For example, see tomy0000000/blog#8

This PR removes all arbitrary spacing and adds a px-1 class to state the spacing explicitly.

### DIFF
--- a/layouts/partials/footer.html
+++ b/layouts/partials/footer.html
@@ -16,9 +16,9 @@
                     <span
                       class="group-dark:hover:text-primary-400 transition-colors group-hover:text-primary-600"
                     >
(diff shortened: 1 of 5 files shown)


jpanther:
Thanks for this fix, @tomy0000000. I'm happy to accept this PR and will merge it in now.

PR #573, opened 2023-06-03
🌐 Update zh-CN i18n

DejavuMoe:
- [x] Updated Simplified Chinese translation, changed `i18n/zh-CN.yml` file


jpanther:
Thanks for this update, @DejavuMoe. I'll merge it in now and it will be included in the next theme release.

PR #575, opened 2023-06-04
Add spiffyeight77.com to user sites page

SpiffyEight77:
This theme design for Hugo is amazing. Thank you very much.


jpanther:
Thanks for using Congo, @SpiffyEight77! I'll merge this change in and your site will appear on the public list when the next theme release is published.

PR #580, opened 2023-06-06
:memo: Tiny typo fix: broswer to browser

major:


jpanther:
Thanks for picking this up!

PR #584, opened 2023-06-16
Add support for Plausible analysis service

DejavuMoe:
**[Plausible](https://plausible.io/) is a simple, privacy-friendly alternative to Google Analytics, in this Pr:**

- [x] Add Plausible analysis script, but not enabled by default
- [x] Add configuration fields for Plausible analysis in `params.toml`
- [x] Support users to use Plausible management deployment or self-hosting services in `params.toml`
- [x] Add the appropriate description to the document


jpanther:
Thanks for the PR, @DejavuMoe! I'm happy to look at adding this to the theme as it's a broadly useful feature. Is it possible to refactor this so that the `enable` config flag is not required? ie. Could it work in a similar way to how Fathom Analytics or Google Analytics works in that when you provide a site ID/domain it then knows to include the code. It would provide for a more consistent user experience.

DejavuMoe:
> Thanks for the PR, @DejavuMoe! I'm happy to look at adding this to the theme as it's a broadly useful feature. Is it possible to refactor this so that the `enable` config flag is not required? ie. Could it work in a similar way to how Fathom Analytics or Google Analytics works in that when you provide a site ID/domain it then knows to include the code. It would provide for a more consistent user experience.

The `enable` field can now be dispensed with, and for services deployed using Plausible's official management, only the domain to be tracked need be provided. 

Specifically, the `event` and `script` fields are retained, which is useful when users use self-hosted Plausible service or proxied Plausible API event and script.

jpanther:
Thanks for the update. The only final change I would make is to rename the `plausible` config block to `plausibleAnalytics`. Again this is to maintain consistency with the existing `googleAnalytics` and `fathomAnalytics` parameters. Once this is done, I'm happy to merge this into the project.

DejavuMoe:
> Thanks for the update. The only final change I would make is to rename the `plausible` config block to `plausibleAnalytics`. Again this is to maintain consistency with the existing `googleAnalytics` and `fathomAnalytics` parameters. Once this is done, I'm happy to merge this into the project.

Now it should be done.

DejavuMoe:
May this PR could be merged? @jpanther 🤔

jpanther:
Thanks again @DejavuMoe. Sorry it's taken me a while to respond. I've been very busy with life challenges! I'm happy to merge this in and it will be included in the next release.

Issue #583, opened 2023-06-16
Better language switcher


jpanther:
### Previous discussion in https://github.com/jpanther/congo/discussions/462

There is a need to overhaul the language switcher and I want to improve it for everyone.

Unfortunately, I have been very busy in my day job and my workload has been too large to deal with this issue. I definitely want to fix it though and there's been a number of challenges for me to get there. As I think I've mentioned elsewhere, my intention is to make the language picker portable so that it can be placed in any menu. The changes I've been making over the last few theme releases have laid the groundwork for this however I'm still impeded by the time constraints of developing the last few components.

One of the biggest hurdles I have is actually having demo data to be able to test new i18n features. I don't have multilingual content in the example site and so I have to make a local development environment every time and set up dummy data for other languages. It takes a lot of time and doesn't reflect real use cases. This means that for me to be able to make changes, it's not as quick as just being able to test it against the large dataset of the current example site. It would be super helpful if the community could chip in here and perhaps we could get some of the sample content in the theme localised so that the example site has some multilingual content by default. It would not only help to demonstrate the theme features to new users, but it would greatly speed up the development and testing process. Having "real" data highlights potential issues far earlier than it does if I'm just creating shell pages to see if the translation features work.

The other issue is there are a lot of opinions on how this feature should operate. Some like the idea of a dropdown language picker, others like icons, some want it to be a top level menu and others want it buried in the hamburger menu. Implementing the feature in such a way that it is customisable is a key component but it's going to be difficult to please everyone. The mobile menu was exactly like this and it took time to come up with a workable solution that meets most people's needs.

I think the change I made to hide the current language wasn't thought out enough, and I expected to have more time to be able to quickly release incremental updates to work towards a longer term solution but I just haven't had the time. My options at this point are to roll it back or push forward knowing that it will take some time to get to the final feature release.

I'm going to keep this issue open as the main place to discuss this change as it will help keep me focused on implementing a solution. I'll link existing discussions to this page. Please post comments and feedback as it really helps me understand everyone's needs and wants.

AntoineSoetewey:
Dear @jpanther,

Thanks for your reply. I totally understand that it takes time and your workload doesn’t allow to tackle this issue.

In the meantime, here are some thoughts regarding its position:

As far as I know, one person puts the language switcher in the footer (see https://github.com/jpanther/congo/discussions/290), while most people would rather put it in the menu (see https://github.com/jpanther/congo/discussions/462) and it seems the case for most multilingual websites.

[...]

PR #587, opened 2023-06-23
🗄 chore: Add Tomy's Blog

tomy0000000:
Adding my blog to the user's site page.

I have a custom translation dropdown menu at the footer, which might be something you're looking for. I'll elaborate more in #583.


jpanther:
Thanks for using Congo, @tomy0000000! I'll merge this change in and your site will appear on the public list when the next theme release is published.

PR #605, opened 2023-07-15
🩹 Replace deprecated shortened URL

sorairolake:
Git.io is [deprecated](https://github.blog/changelog/2022-04-25-git-io-deprecation/), so replace this.


jpanther:
Thanks, I hadn't released this service had been discontinued. Happy to merge this change in.

PR #606, opened 2023-07-16
🌐 Translation of the example site into Spanish

alejandrgaspar:
Hello there!

As discussed in issue #583, one of the main challenges in developing new i18n features is the lack of multilingual content on the example site. So, I took the liberty of translating the entire example site into Spanish (except for the documentation part, which, will require a lot of effort from the community).

Just like @jpanther has mentioned in the same issue, the example site is already structured in subdirectories. That's why the changes I'm proposing simply involve adding translated versions of the existing markdown files, with a `.es.md` extension.

It would be great if another Spanish native speaker could take a look at the translations since it's not my area of expertise. I believe they serve the purpose well and are of decent quality, but it would still be a nice touch. I did my best to ensure a neutral translation, but I might have missed something. Any assistance would be appreciated!

I look forward to your comments.


jpanther:
Thanks for the contribution, @alejandrgaspar! As I don't know Spanish I'll just have to merge this in as is and if others have comments once the pages are live on the demo site, then it can be adjusted as necessary.

PR #608, opened 2023-07-17
✏️ Fix typo in users list page

alejandrgaspar:
While reading the example site to do the Spanish translation, I noticed this typo.


jpanther:
Good find! Thanks for fixing this up.

PR #609, opened 2023-07-17
🩹 Update page source URL

alejandrgaspar:
On the Charts and Diagrams pages the links to view the source code are outdated. So modify the URLs to match the current state of the project.

Btw: sorry for the spam 😅.


jpanther:
Another good find. Thanks again.

PR #615, opened 2023-07-22
Added simaosilva.com

Simons36:
Added https://simaosilva.com to the users list.

Thanks for the amazing theme!


jpanther:
Thanks for using Congo, @Simons36! I'll merge this change in and your site will appear on the public list when the next theme release is published.

PR #630, opened 2023-08-16
Use correct render value in external archetype

alexmufatti:
On hugo documentation render params should have value "render" not "false".
Also removed the list param to allow external to be shown in homepage.


jpanther:
Thanks, I'm happy to merge this change in.

PR #645, opened 2023-09-11
Provide width and height for images

stereobooster:
Provide `width` and `height` for images to reduce [layout shift](https://web.dev/cls/).

As proposed here https://github.com/jpanther/congo/discussions/644


jpanther:
Thanks, this change makes sense and I'm happy to merge it in. I think that image dimensions were in the theme many releases ago but it was removed in a refactor and never picked up until now.

PR #646, opened 2023-09-11
Add quicklink

stereobooster:
Add quicklink, as proposed here https://github.com/jpanther/congo/discussions/644

Related Firefox issue: https://github.com/GoogleChromeLabs/quicklink/issues/262#issuecomment-1479681487


jpanther:
I do wonder how much of a performance benefit this adds to a static site, but I'm not opposed to adding it in. Is it worth limiting how many links can be prefetched?

stereobooster:
> I do wonder how much of a performance benefit this adds to a static site,

Depends how you judge it. Let's say that fastest server (CDN) can deliver HTTP response in 100-200ms (mainly network, I guess server itself would take 1ms max to generate response). It seems like not much, but page navigation (at least to my perception) becomes instant, like in SPA.

> Is it worth limiting how many links can be prefetched?

Not sure. `quicklink` already uses some heuristics, like it preloads only visible links (in viewport)

jpanther:
Yeah I think this change is not going to cause many issues for the majority of users and if anyone does have a problem it's easy to opt-out with the config flag so I'm happy to merge it in.

PR #647, opened 2023-09-11
LQIP proposal

stereobooster:
**Matter of taste**

---

As proposed here https://github.com/jpanther/congo/discussions/644. Requires https://github.com/jpanther/congo/pull/649 PR first. We may experiment with [blur](https://developer.mozilla.org/en-US/docs/Web/CSS/filter-function/blur) or [backdrop-filter](https://developer.mozilla.org/en-US/docs/Web/CSS/backdrop-filter)

Loading on slow network:

Screenshot 2023-09-11 at 10 46 32

Screenshot 2023-09-11 at 10 46 35

Related Firefox issue: https://bugzilla.mozilla.org/show_bug.cgi?id=1556156


jpanther:
Is this a big issue given we're already optimising the images? Have you seen performance problems in deployment with this?

stereobooster:
I don't have strong opinion. Only one use case I can come up with if you have images below the cut and user lost internet connection, so when they gonna scroll they will see empty rectangles, but with lqip they will see at least something

**UPD**: On slow networks this still can make sense even if we optimize images

stereobooster:
With blur it looks much better

Screenshot 2023-09-13 at 09 39 59
Screenshot 2023-09-13 at 09 40 17

PR #648, opened 2023-09-11
Applied capo recommendations

stereobooster:
**Matter of taste**. I understand if you don't want to mess up with source-code in order to obey Capo recommendations

---

Applied capo recommendations, as proposed here https://github.com/jpanther/congo/discussions/644

Before:
Screenshot 2023-09-11 at 11 08 16

After:
Screenshot 2023-09-11 at 11 11 42

### DIFF
--- a/exampleSite/config/_default/config.toml
+++ b/exampleSite/config/_default/config.toml
@@ -4,6 +4,7 @@
 
 theme = "congo"
 defaultContentLanguage = "en"
+disableHugoGeneratorInject = true
 
 enableRobotsTXT = true
 paginate = 15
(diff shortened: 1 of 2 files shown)


jpanther:
Hmm I've never heard of this capo project before. I wonder how much this actually impacts performance? I don't think any of these types of changes were flagged by Lighthouse or GTmetrix although I haven't run those tests formally in a little while.

stereobooster:
No they are not flagged by Lighthouse. I agree this may be too much. I'm myself wasn't able to see the difference.

stereobooster:
I will close it for now. We can reopen it, if needed

PR #650, opened 2023-09-12
Small improvement in js

stereobooster:
it is more memory efficient to do one `join`,  than accumulate string


jpanther:
I'm getting a TypeError when running this. Are you using this change in a production environment?

stereobooster:
Sorry. I made a mistake: it suppose to be `map`, not `forEach`

jpanther:
Thanks, I'm happy to accept this change now. I'll merge it in.

PR #654, opened 2023-09-17
Update index.md

knowkomal:
Personal site using congo - kom.al


jpanther:
Thanks for using Congo, @knowkomal! I'll merge this change in and your site will appear on the public list when the next theme release is published.

PR #655, opened 2023-09-19
Portable markdown links

stereobooster:
See [Portable markdown links](https://stereobooster.com/posts/portable-markdown-links/)

### DIFF
--- a/layouts/_default/_markup/render-link.html
+++ b/layouts/_default/_markup/render-link.html
@@ -1,7 +1,10 @@
-<a
-  href="{{ .Destination | safeURL }}"
-  {{ with .Title }}title="{{ . }}"{{ end }}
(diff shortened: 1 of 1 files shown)


jpanther:
I don't see any harm in making this change. It adds functionality while maintaining the existing workflow. I only note that you have omitted the `rel="noreferrer noopener"` part of the link and I believe this is still required? It was added in https://github.com/jpanther/congo/pull/446 but potentially this is no longer necessary?

stereobooster:
Anchor `target=_blank` implies `rel=noopener` by default ([as of 2021, for most browsers](https://chromestatus.com/feature/6140064063029248)). But not `noreferrer`. So yes, I need to fix this bit

jpanther:
Thanks for amending it, I'll merge this one in now.

PR #658, opened 2023-09-24
Add page context to the extend-head feature

### LINKED ISSUE #657: Cannot access page context when using extend-head feature
### Issue description

I am incorporating plotly.js graphs into my website, which requires a couple of additional `<script>` tags in the `<head>` of the document. 

I found the congo theme feature that allows extending congo's `head.html` partial via the `extend-head.html` file. Perfect! 
However, I also want to add a piece of front matter config to an individual page such that the additional scripts are only loaded on pages that need it (when `plotly: true` is specified in the front matter of a given page). This functionality requires the following code to be placed in `extend-head.html`
```html
{{ if .Params.plotly }}
<script src="https://d3js.org/d3.v7.min.js"></script>
<script src="https://cdn.plot.ly/plotly-2.11.1.min.js"></script>
{{ end }}
```

When I build the site, the plots won't render, but there are no errors. If I remove the conditional logic, then the plots do render, but the scripts are included in all pages including ones that don't need them.

After some digging, I found that when congos `head.html` partial calls the `extend-head.html` partial, it passes `.Site` as the context, which prevents accessing the page front matter. The issue is fixed when I instead pass `.` as the context to `extend-head.html`.

I am opening this as an issue because I think it's a bug, but I can't rule out the possibility of there being other considerations for passing `.Site` as context. I will note that the sibling partial `extend-footer.html` does pass the full `.` context when called.

### Example

[code](https://github.com/micah-svenson/congo/tree/bug-repro)

To run the example:
1. clone branch and build example site with `npm run example`
```
git clone --branch bug-repro --single-branch https://github.com/micah-svenson/congo.git
```
2. navigate to `http://localhost:8008/congo/samples/plotly/`

The plot won't render until you change the context of the `extend-head.html` call from `.Site` to `.` as the post mentions.
image

### Theme version

2.6.1

### Hugo version

hugo v0.118.2-da7983ac4b94d97d776d7c2405040de97e95c03d+extended darwin/amd64 BuildDate=2023-08-31T11:23:51Z VendorInfo=brew

WebKit (Safari)

### URL to sample repository or website

https://github.com/micah-svenson/congo/tree/bug-repro

### Hugo output or build error messages

_No response_

micah-svenson:
Fixes #657

### DIFF
--- a/layouts/partials/head.html
+++ b/layouts/partials/head.html
@@ -107,6 +107,6 @@
   {{ partial "analytics.html" . }}
   {{/* Extend head - eg. for custom analytics scripts, etc. */}}
   {{ if templates.Exists "partials/extend-head.html" }}
-    {{ partial "extend-head.html" .Site }}
+    {{ partial "extend-head.html" . }}
   {{ end }}
 </head>


jpanther:
Change accepted. This will align the existing behaviour of the extended footer partial and makes the overall theme more flexible.

PR #669, opened 2023-10-15
Add Coffee and X Icons from FontAwesome

s00500:
Was missing these, did not use gitmoj on commits, but can be squashed on merge :-)


jpanther:
Thanks for these additions. I'll merge them in.

s00500:
Awesome, thanks

PR #671, opened 2023-10-17
Fixed taxonomies display issue

bboykarlito:
When there are a lot of tags for the article they begin to be carried to the next line that cause some styling issues. Here is an example:
Before

I suggest to make them look little bit better:
After

There is a real example of this issue occurrence as well. It happened for the [article ](https://ruby.mobidev.biz/posts/faq-chatbot/) in the our developers team`s blog.


jpanther:
Thanks for this fix, @bboykarlito. I'm happy to merge this in.

bboykarlito:
Thank you! We are really enjoy to use this theme!)

PR #680, opened 2023-10-24
:children_crossing: :adhesive_bandage: [SVG_Support: Logo Image] 

wolfspyre:
Add logic brace around img dimension math, so as to support SVGs. 

Add warning discouraging specifying the same asset as both logo and logo_dark.

### DIFF
--- a/layouts/partials/functions/warnings.html
+++ b/layouts/partials/functions/warnings.html
@@ -7,3 +7,6 @@
 {{ if ne .Params.logo nil }}
   {{ warnf "[CONGO] Theme parameter `logo` has been renamed to `header.logo`. Please update your site configuration." }}
 {{ end }}
+{{ if eq .Params.header.logo .Params.header.logo_dark }}
+  {{-  warnf "[CONGO] Interactions with a single resource via multiple distinct resource handles can be problematic. It is recommended to only declare `header.logo_dark` when it is distinct from `header.logo`." }}
+{{- end }}
--- a/layouts/partials/logo.html
+++ b/layouts/partials/logo.html
@@ -5,17 +5,29 @@
     <a href="{{ "" | relLangURL }}" class="mr-2">
       <img
         src="{{ $logo.RelPermalink }}"
-        width="{{ div $logo.Width 2 }}"
-        height="{{ div $logo.Height 2 }}"
-        class="max-h-[10rem] max-w-[10rem] object-scale-down object-left{{ if $logo_dark }} hidden dark:flex{{ end }}"
+        {{- if eq $logo.MediaType.SubType "svg" }}
+          width="25"
+          height="25"
+          class="m-0 p-0 object-scale-down object-left{{ if $logo_dark }} hidden dark:flex{{ end }}"
+        {{- else }}  
+          width="{{ div $logo.Width 2 }}"
+          height="{{ div $logo.Height 2 }}"
+          class="max-h-[10rem] max-w-[10rem] object-scale-down object-left{{ if $logo_dark }} hidden dark:flex{{ end }}"
+        {{- end }}  
         alt="{{ .Site.Title }}"
       />
       {{- if $logo_dark }}
       <img
         src="{{ $logo_dark.RelPermalink }}"
-        width="{{ div $logo_dark.Width 2 }}"
-        height="{{ div $logo_dark.Height 2 }}"
-        class="max-h-[10rem] max-w-[10rem] object-scale-down object-left dark:hidden"
+        {{- if eq $logo_dark.MediaType.SubType "svg" }}
+          width="25"
+          height="25"
+          class="m-0 p-0 object-scale-down object-left{{ if $logo }} !dark:hidden{{ end }}"
+        {{- else }}
+          width="{{ div $logo_dark.Width 2 }}"
+          height="{{ div $logo_dark.Height 2 }}"
+          class="max-h-[10rem] max-w-[10rem] object-scale-down object-left{{ if $logo }} !dark:hidden{{ end }}"
+        {{- end }}
         alt="{{ .Site.Title }}"
       />
       {{- end}}


jpanther:
I think this is a good change, I'm just wondering what hard coding the width and height to 25 for SVGs achieves? Doesn't this constrain the image to a tiny area?

wolfspyre:
I thought the goal was to have the logo entity reside within the height of the menu element, which is ~26px so constraining to 25x25 to avoid being surprised

wolfspyre:
image

jpanther:
The logo supports any image dimensions as some users prefer to hide the site name and use only a logo, in which case they provide large assets and the rest of the menu is laid out accordingly. It would be odd constraining the SVG but allowing any size for other assets.

wolfspyre:
ah! `k.. (the examples I saw when perusing the discussions all seemed to use it as a menu element. mah bad)

I could see the options of:
-  adding optional w/h keys to the header param element, to allow the user to specify the size if they wanted
```
[header]
  layout = "basic" # valid options: basic, hamburger, hybrid, custom
  # logo = "img/logo.jpg"
  # logoDark = "img/dark-logo.jpg"
  # logoWidth =
  # logoHeight =  
```
- removing the specification of dimensions entirely for svg,  delegating sizing to the viewbox element of the svg

got a better idea?

wolfspyre:
I played around with the svg logo a bit today while working on a dropdown menu implementation.

I'm open to feedback WRT how you'd like this improved; or what concerns you have.

Currently Hugo's image `Width` and `Height` methods[^1]  are not able to assess the dimensions of an SVG. 
Resultantly, the current `{{ div $logo.Width 2 }}` and `{{ div $logo.Height 2 }}` operations will need some massaging to work with SVGs.

as it sits, I see a couple options... if you see more, by all means, lets hear 'em.

- hardcode x/y
This is, admittedly, not the most robust solution. Just the simplest.

- user-controlled dimension knobs:  
    provide param keys for `logoWidth` and `logoHeight`; and use them to populate the img `width` and `height` properties in the logo.[^2]  and just follow the current logic of resizing raster images proportionally[^3]

Going much further than this will quickly tread into scope creep / yak shaving territory, as robustifying and do-what-I-mean-nerfing the user-facing knobs has a plethora of nuance[^4]:

My gut says the most helpful thing to do is to simply emit a warning if AND[^5]:
- Both X and Y params have a user-provided value
- the image is not a vector.

With something like:
> You provided values for both `Params.Header.logoWidth` and `Params.Header.logoHeight` and the logo asset is NOT a vector. 
> If the generated image is distorted or misshapen, we advise setting only one of the two dimensions in your configuration

...And then  just continuing to do what the user asked us to do.

Combing this behavior with clear documentation/instructions on usage. AUGHT be aligned with most users' expectations, I THINK? 

This does kinda set the stage to start using the new image pipeline stuff,  which has a bunch of new fun features.... but that's all out of scope for this simple change.... 

thots?
 
[^1]: at least on 0.119
[^2]: Default would be to leave them at `0`, `-1` or `auto` or something....
[^3]: May want to also automagically adjust the max-h-[Y] / max-w-[X] classes associated with the logo img... but maybe not
[^4]: If you only provide one dimension axis to Hugo's resizing tools, it's smart about retaining proportions... 
if you provide both X and Y, it will dutifully scale the image as you requested; even if that would contort the image... which is LIKELY not what most users would expect or desire.
[^5]: If its simple enough, perhaps only emit this warning if using the x/y provided would result in a change to the aspect ratio of the image.  ¯\\\_(ツ)_/¯

wolfspyre:
I came up with a different, mechanism you might like a lil more?

`layouts/partials/logo.html`:
```gotmpl
{{- if .Site.Params.header.logo }}
{{-   $logo := resources.Get .Site.Params.header.logo.source }}
{{-   if $logo }}
{{-     $w := -1 }}{{ if ne $logo.MediaType.SubType "svg"}}{{ $w = (div $logo.Width 2) }}{{ end }}
{{-     $h := -1 }}{{ if ne $logo.MediaType.SubType "svg"}}{{ $h = (div $logo.Height 2) }}{{ end }}
{{-     $logoWidth  := .Site.Params.header.logo.width | default  $w  }}
{{-     $logoHeight := .Site.Params.header.logo.height| default  $h }}
{{-     $logoDark := resources.Get .Site.Params.header.logoDark.source }}
    <a href="{{ "" | relLangURL }}" class="mr-2">
      <img
        src="{{ $logo.RelPermalink }}"
        width="{{ $logoWidth }}"
        height="{{ $logoHeight }}"
        class="max-h-[10rem] max-w-[10rem] object-left {{ if $logoDark }} flex dark:hidden {{ end }}"
        alt="{{ .Site.Title }}"
      />
{{-     if $logoDark }}
{{-       $dw := -1 }}{{ if ne $logoDark.MediaType.SubType "svg" }}{{ $dw = (div $logoDark.Width 2)  }}{{ end }}
{{-       $dh := -1 }}{{ if ne $logoDark.MediaType.SubType "svg" }}{{ $dh = (div $logoDark.Height 2) }}{{ end }}
{{-       $logoDarkWidth  := .Site.Params.header.logoDark.width | default $dw }}
{{-       $logoDarkHeight := .Site.Params.header.logoDark.height| default  $dh }}
          <img
        src="{{ $logoDark.RelPermalink }}"
        width="{{ $logoDarkWidth  }}"
        height="{{ $logoDarkHeight }}"
        class="max-h-[10rem] max-w-[10rem] object-scale-down object-left hidden dark:flex"
        alt="{{ .Site.Title }}"
      />
{{-     end}}
    </a>
{{-   end }}
{{- end }}
```
`config/_default/params.yaml`:
```yaml
header:
  layout: 'grace' # hybrid, hamburger, basic.
  showTitle: false
  logo: 
    source: '/svg/Atticus-in-square-frame.svg'
    width: 100
    height: 100
  logoDark: 
    source: '/Wolfspyre_Logo.svg'
    width: 100
    height: 100
  showAppearanceSwitcher: true
```

lightsvg
darksvg

you like this more? I can update the PR; but I'd want to tweak a bit more, as currently this breaks existing behavior, since it converts the struct from a string to a dict.. I'd need to wire in support for what exists today (string as $logo/$logoDark) so it's still backwards compatible, and update doc to say that this is how one would support svg's... but turning the image construct from just a bare string to an object allows us to bolt in other stuff later (ie, output format/ image type to emit, pipeline operations to perform,  caption, description, src_url for remote assets, etc ) if we want to... ¯\\\_(ツ)_/¯

wolfspyre:
for the record I'm not even sure *I* like this more, just trying to come up with other ways to solve the problem....

I was **INITIALLY** crazily hoping that if I added .Width/.Height as params of the image, that I could sidestep the .Width/.Height methods not working on vectors problem... but that was a stroke of insanity, not genius... however since I was already halfway thru banging it together, I figured I'd at least put it together enough to see if you like this approach more.

PR #691, opened 2023-10-31
Print warning if system can't resolve link to a markdown file

stereobooster:
Small improvement:

if user uses link, like `[post](/content/posts/other-post.md)` and this file doesn't exist (e.g. broken link), Hugo will print a warning. Which is nice way to prevent 404 errors

### DIFF
--- a/layouts/_default/_markup/render-link.html
+++ b/layouts/_default/_markup/render-link.html
@@ -2,10 +2,16 @@
 {{ $isRemote := strings.HasPrefix $link "http" }}
 {{- if not $isRemote }}
   {{ $url := urls.Parse .Destination }}
-  {{ if $url.Path }}
+  {{- if $url.Path }}
     {{ $fragment := "" }}
     {{- with $url.Fragment }}{{ $fragment = printf "#%s" . }}{{ end -}}
-    {{- with .Page.GetPage $url.Path }}{{ $link = printf "%s%s" .RelPermalink $fragment }}{{ end -}}
-  {{ end }}
+    {{- with .Page.GetPage $url.Path }}
+      {{ $link = printf "%s%s" .RelPermalink $fragment }}
+    {{ else }}
+      {{- if hasSuffix $url.Path ".md" }}
+        {{ warnf "Can't resolve: %s" .Destination }}
+      {{ end -}}
+    {{ end -}}
+  {{ end -}}
 {{ end -}}
 <a href="{{ $link | safeURL }}"{{ with .Title }} title="{{ . }}"{{ end }}{{ if $isRemote }} target="_blank" rel="noreferrer"{{ end }}>{{- .Text | safeHTML -}}</a>


wolfspyre:
wouldn't  also emitting the page this occurs at by using `.Page.Path`: 
`{{-    warnf "DEBUG We are disabled, but were invoked by  %s " .Page.Path -}}`

be helpful? 

also, might wanna have this be a toggle like: (totes written off the cuff and prolly wont work as is)

```
{{- $.Scratch.Set "linkLiveness"  "" }}
{{- $.Scratch.Set "deadLinkErr"  false }}
{{- $linkLiveness := (.Site.Params.Warnings.LinkLiveness | default false)  }}
{{- if not (in (slice "warn" "error" "disabled")) $linkLiveness }}
{{-   $linkLiveness = "disabled" }}
{{- end }}
{{- $.Scratch.Set "linkLiveness" $linkLiveness }}
```
... pseudocode that I'm sure won't work
```
{{- with {dead link path} }}
{{-    $.Scratch.Set "deadLinkErr"  (printf "dead link on page %s to %s" .Page.Path .) }}
{{- end }}
...
{{- if and   $.Scratch.Get "linkLiveness" (eq ($.Scratch.Get "linkLiveness") "error" ) -}}
{{-   errorf "%s" $.Scratch.Get "deadLinkErr" -}}
{{- else if and   $.Scratch.Get "linkLiveness" (eq ($.Scratch.Get "linkLiveness") "warn" ) -}}
{{-  /* warn dont fail */ -}}
{{-   warnf "%s" $.Scratch.Get "deadLinkErr" -}}
{{- else  -}}
...
{{- end  -}}
```

or you could use the errorf variant that has an excludable errnumber mechanism that allows the user to ignore errors of type _ ....

jpanther:
Looking at the code for this warning, doesn't this prevent any links to `*.md` URLs? What happens if the user is wanting to link to a Markdown file?

stereobooster:
> Looking at the code for this warning, doesn't this prevent any links to *.md URLs? What happens if the user is wanting to link to a Markdown file?

Yes it will print warning too. It would be something like this:

```
/posts/slug-with.md
```

e.g. if slug ends with `.md`. Not sure this can be fixed somehow. Related https://github.com/gohugoio/hugo/issues/10585

PS it will not print warning for external links, like

```
https://github.com/jpanther/congo/blob/dev/README.md
```

jpanther:
The only that's stopping me from merging this is that there may be edge cases that will be triggered here that we're missing. And I wonder if throwing warnings when people want to link to Markdown files directly is an issue. I'm not sure if this is a prevalent use case or not.

stereobooster:
Hard to tell. It will break nothing. People will see some warnings which doesn't stop them from publishing it as is anyway. We can implement it and see if somebody complains or gets confused by it.

Other way would be to add config which would disable this check in case somebody gets a lot of false warnings.

wolfspyre:
if that’s your only reservation, perhaps changing the verbiage of the warning to indicate a potential red herring or a reference to this PR with a request to comment if an edge case pops up, and leave it as a potential nit for FutureUs™️? and then fix forward if/when it bites us in the butt as we’ll have a better sense of what edge case arose at that point?

jpanther:
I'm probably overthinking this one. I'll merge it in and we can address any issues that arise if and as they appear.

PR #692, opened 2023-10-31
Bump js libraries

stereobooster:
I didn't change anything, except I ran

```
npm i
```


jpanther:
Your commit downgrades these dependencies! I don't think you're running the latest version of the codebase when you're invoking `npm`. Closing this issue as dependabot keeps dependencies up to date.

stereobooster:
> I don't think you're running the latest version of the codebase when you're invoking npm

strange. I pulled latest content from upstream. But maybe something is wrong in my setup

stereobooster:
But https://github.com/jpanther/congo/blob/54c39a1297faeb2a44eb5d3d18a246e7197484e4/package-lock.json#L3 version in package-lock is not bumped 

But if I run `npm ci` everything is ok

jpanther:
That is more a side effect of the release process than an issue with dependencies as the package version is updated on deploy but I'm not specifically running `npm install` at that time which means the package-lock is always a version behind. But this shouldn't cause any issues when it comes to running a local build.

Issue #600, opened 2023-07-12
Images flipped 90 degrees


tomleary:
### Issue description

Images taken horizontally show up fine, but anything taken vertically is rendered horizontally.
<img width="648" alt="Screenshot 2023-07-12 at 3 57 40 PM" src="https://github.com/jpanther/congo/assets/82277734/b37200c0-bce7-4e41-b265-caa366a569df">

The code is simply:
```
![Alt text](ghat3.jpg "Image caption")
![Alt text](ghat4.jpg "Image caption")
```
I feel like this can't be standard behavior, so maybe I accidentally changed a configuration file without realizing it. Otherwise, it might have to do with the way the theme tries to make all the images the same width.

### Theme version

v2.6.1

### Hugo version

hugo v0.112.1+extended

### Which browser rendering engines are you seeing the problem on?

Chromium

DejavuMoe:
Are you taking pictures with iPhone, which seems to have incorrectly rotated EXIF information?

tomleary:
Those photos were taken with an Android, though I've had the same problem with iPhone photos, too. All the photos show up fine in every other image-viewing application. They also show up fine if I embed them with HTML instead of Markdown or the figure shortcode.
<img width="624" alt="Screenshot 2023-07-12 at 8 56 21 PM" src="https://github.com/jpanther/congo/assets/82277734/45a294ae-727f-4940-b6a7-4af6f17e5987">
```
<img src="ghat3.jpg" />
<img src="ghat4.jpg" />
```

BenAlexanderAU:
To assist (as a workaround and troubleshooting)...

I also have this issue with most (but not all) portrait images taken on my Android phone.
As a workaround, I am using an image editor on my PC to rotate the image 90 degrees, save, and then rotate back and save again... Now the image will appear in the correct orientation on the website.

I believe this indicates it is an EXIF issue. I am uncertain if the Android EXIF data needs to be fixed, or if there is another orientation indicator Huge can use.

jpanther:
I have seen this commented from time to time in the discussions and there's a lot on the Hugo Discourse about it too. There seems to be issues with how EXIF data is interpreted and that Hugo doesn't rotate the images to match the EXIF orientation. I might have to see if I can add this logic to the theme itself. See https://discourse.gohugo.io/t/image-exif-orientation/22902 for details.

jmooring:
<https://gohugo.io/functions/images/autoorient/>

[...]

PR #693, opened 2023-10-31
feat: picture partial

stereobooster:
Example to demonstrate what is discussed here https://github.com/jpanther/congo/pull/649#issuecomment-1780335545

Example of generated code:

```html
<figure class="m-auto mt-6 max-w-prose">
  <picture class="m-auto mt-6 max-w-prose">
    <img src="/congo/festivities.svg" width="1086" height="811.37" class="m-auto mt-6 max-w-prose" loading="lazy" decoding="async">
  </picture>
</figure>
```

It can detect width and height for SVGs. There is branch for LQIP in code, but it's not used now.

### DIFF
--- a/layouts/_default/_markup/render-image.html
+++ b/layouts/_default/_markup/render-image.html
@@ -12,7 +12,7 @@
 
 {{/* https://github.com/gohugoio/hugo/pull/10666/files */}}
 {{- $params := $url.Query -}}
-{{- $x2Param := $params.Get "x2" -}}
+{{- $x2Param := $params.Get "2x" -}}
 {{- $x2 := false -}}
 {{- if eq $x2Param "true" -}}
     {{- $x2 = true  -}}
--- a/layouts/_default/_markup/render-image.html
+++ b/layouts/_default/_markup/render-image.html
@@ -10,7 +10,7 @@
     {{ $img = resources.Get $path }}
 {{ end -}}
 
-{{/* https://github.com/gohugoio/hugo/pull/10666/files */}}
+{{/* https://github.com/gohugoio/hugo/pull/10666 */}}
 {{- $params := $url.Query -}}
 {{- $x2Param := $params.Get "2x" -}}
 {{- $x2 := false -}}
--- a/layouts/_default/_markup/render-image.html
+++ b/layouts/_default/_markup/render-image.html
@@ -20,7 +20,7 @@
 
 <figure>
   {{- with $img -}}
-    {{ partial "pictureDefaults.html" (dict "img" . "alt" $altText "class" $class "x2" $x2) }}
+    {{ partial "picture.html" (dict "img" . "alt" $altText "class" $class "x2" $x2) }}
   {{- else -}}
     <img src="{{ .Destination | safeURL }}" alt="{{ $altText }}" class="{{ $class }}"/>
   {{- end -}}
--- a/layouts/_default/_markup/render-image.html
+++ b/layouts/_default/_markup/render-image.html
@@ -10,9 +10,17 @@
     {{ $img = resources.Get $path }}
 {{ end -}}
 
+{{/* https://github.com/gohugoio/hugo/pull/10666/files */}}
+{{- $params := $url.Query -}}
+{{- $x2Param := $params.Get "x2" -}}
+{{- $x2 := false -}}
+{{- if eq $x2Param "true" -}}
+    {{- $x2 = true  -}}
+{{- end -}}
+
 <figure>
   {{- with $img -}}
-    {{ partial "pictureDefaults.html" (dict "img" . "alt" $altText "class" $class) }}
+    {{ partial "pictureDefaults.html" (dict "img" . "alt" $altText "class" $class "x2" $x2) }}
   {{- else -}}
     <img src="{{ .Destination | safeURL }}" alt="{{ $altText }}" class="{{ $class }}"/>
   {{- end -}}
--- a/layouts/_default/_markup/render-image.html
+++ b/layouts/_default/_markup/render-image.html
@@ -1,64 +1,20 @@
 {{ $url := urls.Parse .Destination }}
 {{ $altText := .Text }}
 {{ $caption := .Title }}
-{{ $lazyLoad := $.Page.Site.Params.enableImageLazyLoading | default true }}
-{{ if findRE "^https?" $url.Scheme }}
-  <figure>
-    <img
-      class="mx-auto my-0 rounded-md"
-      src="{{ $url.String }}"
-      alt="{{ $altText }}" 
-      {{ if $lazyLoad }}
-        loading="lazy"
-      {{ end }}
-    />
-    {{ with $caption }}<figcaption class="text-center">{{ . | markdownify }}</figcaption>{{ end }}
-  </figure>
-{{ else }}
-  {{ $resource := "" }}
-  {{ if $.Page.Resources.GetMatch ($url.String) }}
-    {{ $resource = $.Page.Resources.GetMatch ($url.String) }}
-  {{ else if resources.GetMatch ($url.String) }}
-    {{ $resource = resources.Get ($url.String) }}
-  {{ end }}
-  {{ with $resource }}
-    <figure>
-      <img
-        class="mx-auto my-0 rounded-md"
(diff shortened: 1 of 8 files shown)


layouts/_default/_markup/render-image.html
```diff
@@ -1,64 +1,20 @@
...
+{{- if and (not $img) .Page.File }}
+    {{ $path := path.Join .Page.File.Dir $file }}
+    {{ $img = resources.Get $path }}
+{{ end -}}
+
+<figure>
+  {{- with $img -}}
+    {{ partial "pictureDefaults.html" (dict "img" . "alt" $altText "class" $class) }}
+  {{- else -}}
+    <img src="{{ .Destination | safeURL }}" alt="{{ $altText }}" class="{{ $class }}"/>
```
stereobooster, line 25: there is no usage of lazy param for this case, and it's intentional. I think it's better to not use lazy for images without dimensions

wolfspyre:
would it make more sense to include this as a well documented shortcode as initial implementation? 
thus making its' usage optional?

further work would likely involve documentation changes to point out the integration as it gets rolled into default page templates?

stereobooster:
> would it make more sense to include this as a well documented shortcode as initial implementation?
thus making its' usage optional?

Which scenario do you have in mind? This partial just replaces similar functionality which is already there (in image render hook, figure shortcode and "single" template). Can't see how it would be useful as shortcode

wolfspyre:
the thought was:

introduce the new component in a way that it can be evaluated without forcing it to be used.. 

this looks like a pretty slick wrapping of figure/img. 

I could see this being introduced as a picture shortcode which invokes your partials.

That would get the partials available, and allow tire-kicking without a hard requirement to use it in the default page templates all in one go....  
IE building comfortability with the new featureset and hunting for weird rough edges or unanticipated side effects 
¯\\\_(ツ)_/¯

stereobooster:
Oh I see. I didn't get the idea from the first message, sorry. Anyway it's up to the author

jpanther:
I've finally had a little more time to have a look at this PR - sorry it's taken a while. Having the option to disable webp in the config is good and I think defaulting it to `true` is the right move. I like that this is removing all the code duplication as this has started to creep into more places than it needs to. Can we add the `screenshot` shortcode in so it uses the same logic too?

I don't see any reason why this couldn't be used across all the templates... it doesn't seem to reduce the functionality in those places and makes everything consistent. The only thing we need to be mindful of is to not lazy load the feature image.

stereobooster:
>  Can we add the screenshot shortcode in so it uses the same logic too?

What about this part?

```
width="100%"
height="auto"
style="max-width:{{ div $image.Width 2 }}px; max-height:{{ div $image.Height 2 }}px;"
```

It is different compared to all other places. Shall I remove it?

wolfspyre:
is that in the screenshot shortcode? it's hard to know without referencing the  file in question

jpanther:
The screenshot shortcode exists mainly because when I post HiDPI screenshots taken on my Mac they would all end up output at double their actual size. This was my way of solving that issue... but maybe there's a better way that doesn't require this?

stereobooster:
> The screenshot shortcode exists mainly because when I post HiDPI screenshots taken on my Mac they would all end up output at double their actual size. This was my way of solving that issue... but maybe there's a better way that doesn't require this?

I think yes - add css class with `max-width: 100%; height: auto`. Which is already the case:

```css
img, video {
  max-width: 100%;
  height: auto;
}
```

browser will calculate ratio based on `width`, `height` from the image and will make image as wide as container. Like any other picture in the content. Does this fit?

jpanther:
The problem here is it makes the screenshots appear blurry when they are scaled up and I like to maintain consistency between the sizes of the screenshots when there are multiple in the same article. Otherwise you end up with some that are larger than others.

I guess I'm just too picky about this one and that's why it exists as its own shortcode. This is one of the only things left in the theme that is selfishly there to solve my own use case. 😆 This is also why it's not referenced in the documentation. Perhaps it doesn't need to exist at all and I could just move it to my own project.

stereobooster:
> The problem here is it makes the screenshots appear blurry when they are scaled up and I like to maintain consistency between the sizes of the screenshots when there are multiple in the same article.

It's fair point. We should not upscale images (at least they should not be blurry). I need to test how it works, for some reason I think it should be ok, because `max-width: 100%;`, not **width**. But let me check

stereobooster:
yes I see what you are talking about

Screenshot

```md
![](lighthouse.jpg)

{{< screenshot src="lighthouse.jpg" >}}
```

I wonder if Hugo has a way to detect DPI and based on that we can adjust to calculation of width, height.

stereobooster:
Wait a second. At first I thought issue is DPI (and browser upscales image), but really problem is in this part:

```
        {{ else }}
          srcset="
          ...
          {{- (.Resize "1320x").RelPermalink }} 2x"
          src="{{ (.Resize "660x").RelPermalink }}"
        {{ end }}
```

Image size is `1200px`, but resized image is `1320px`, so it's not browser. It's the code (Hugo) upscales image. We need to add some kind of condition to not produce images larger than they are.

This problem exists in original code. And I [knew about it](https://github.com/stereobooster/hugo-ideal-image#features)

> - sets HTML attribute srcset and resizes the image
>    - but doesn't upscale image **(not implemented)**

**UPD**: nope it doesn't really fixes the issue, but it is definitely improvement:

Screenshot

jpanther:
I'm not sure simply preventing upscaling solves the issue though. For instance, if I had a 600px wide HiDPI screenshot, that should be displayed at 300px wide in the HTML, not 600px. That's why I divide the dimensions by 2 in the `screenshot` shortcode.

stereobooster:
> I'm not sure simply preventing upscaling solves the issue though. For instance, if I had a 600px wide HiDPI screenshot, that should be displayed at 300px wide in the HTML, not 600px. That's why I divide the dimensions by 2 in the screenshot shortcode.

Yes, you are right. In fact there were both problems. Upscaling in Hugo and upscaling in the browser. I fixed issue with Hugo. And it makes things look better, but not ideal

jpanther:
I don't think the screenshot use case is going to be easy to solve with the existing functions available because there's no easy way to determine if the image is a HiDPI screenshot and not just a photo or some other asset. I think it makes sense to just keep it separate for now. I don't think many people are as pedantic as me when it comes to that particular display issue.

I also wonder if this new partial should try to solve the EXIF rotation issue described in https://github.com/jpanther/congo/issues/600 too? It has been raised a few times by users and is a silly oddity in the way that Hugo handles the EXIF data (that is to say it seems it just ignores the EXIF orientation).

stereobooster:
Here are 2 potential solutions to screenshot issue:

```md
![](lighthouse.jpg?x2=true)

{{< screenshot src="lighthouse.jpg" >}}
```

Screenshot


Give it a try (latest commit in this PR)

stereobooster:
> I also wonder if this new partial should try to solve the EXIF rotation issue described in https://github.com/jpanther/congo/issues/600 too?

Sounds like a separate PR to me

jpanther:
> ```
> 
> ```

I think this is a good solution and it makes everything uniform across the various image types. It also allows me to get rid of the `screenshot` shortcode. The only thing I would change is to flip the parameter name around to be `2x` as those assets are often named using the `@2x` suffix.

stereobooster:
If use code like this:

```liquid
{{ $2x := .2x | default false }}
```

Hugo complains: 

```
 parse failed: template: partials/picture.html:7: bad number syntax: ".2x"
```

I can change it in query params though. Or we can come up with different name for internal variables

stereobooster:
I didn't remove `screenshot` shortcode - don't want to blow up size of PR. Can do a separate PR

jpanther:
This is looking good now. I'll merge it in at this point and we can make future adjustments as necessary.

I think the next logical step is to remove the `screenshot` shortcode and we just bundle it all into `figure` with the `2x` param. It makes no sense keeping it as two different shortcodes because you could have screenshots that aren't HiDPI and in that case it makes no sense. Much easier to just let the user specify what they want and keep it all uniform.

jpanther:
Thanks for your work on this PR @stereobooster, it's much appreciated! This is a great refactor and it removed a whole bunch of code duplication.

PR #694, opened 2023-11-02
Add icon for the threads social network

ned21:
Import FontAwesome icon v6.4.2 for 'threads' and add it to list of all available icons on the exampleSite.


jpanther:
Thanks for this addition, @ned21, and thanks for updating all the docs pages too! I'll merge this one straight in.

Issue #678, opened 2023-10-24
Empty block when locale switcher is hidden


tomy0000000:
### Issue description

Following the recent change in #583, the locale switcher will not be rendered if the page has only one language available. However, the menu still generates an empty list item tag, causing an empty block that doesn't look right.

![CleanShot 2023-10-24 at 06 39 14@2x](https://github.com/jpanther/congo/assets/23290356/490d09b8-959b-4f4e-bc12-1499fab7e300)

![CleanShot 2023-10-24 at 06 39 20@2x](https://github.com/jpanther/congo/assets/23290356/2fb9eef7-21db-4511-b7da-5b01e1ad69bd)

### Theme version

2.7.0

### Hugo version

0.119.0

### Which browser rendering engines are you seeing the problem on?

Chromium (Google Chrome, Microsoft Edge, Brave, Vivaldi, Opera, etc.)

tomy0000000:
The same issue also occurs in the footer.

https://github.com/jpanther/congo/blob/edbb9c0b7485f2be67ae263970598779f3c45f2c/layouts/partials/footer.html#L101-L131

When both the Locale switcher and Appearance switcher are hidden, this creates an empty div.

jpanther:
Just so I'm correct in my understanding here... the issue occurs specifically if the page has no translations, but the menu config still includes the `locale` link? I can see how that is happening as the menu is iterated before the translations partial is included and so the list item is output empty. I just want to check I'm not missing other edge cases that you might be seeing.

tomy0000000:
> Just so I'm correct in my understanding here... the issue occurs specifically if the page has no translations, but the menu config still includes the `locale` link? I can see how that is happening as the menu is iterated before the translations partial is included and so the list item is output empty. I just want to check I'm not missing other edge cases that you might be seeing.

From what I know, yes.

jpanther:
Okay, I'll make some adjustments to fix that case, although I think this is now getting into the territory where the headers need to be completely refactored to simplify some of the logic, but that's for another day.

tomy0000000:
The fix is on the right track but slightly off. I'm drafting a PR to fix this...

PR #699, opened 2023-11-08
🛠 fix: Empty block when locale switcher is hidden (#678)

tomy0000000:
Minor fix for #678


jpanther:
Thanks for picking this one up, @tomy0000000. I'll merge it in now.

PR #705, opened 2023-11-12
fix: close menu button on click

### LINKED ISSUE #460: Menu is not closing for one-page site
### What happened?

Hello,

my website www.datanalyze.be is a one-page site in the sense that links in the menu redirects to a section of the home page.

I noticed that when I click on a link in the menu, it correctly redirects to the section but the menu stays open (and we have to click on the cross to close it).

See in practice: 

https://user-images.githubusercontent.com/17910063/215342929-5b8ce2e9-cbf0-4525-88e5-160316f08399.MOV

Would it be possible to implement the fact that the menu closes automatically after clicking on a link?

In case it helps, the repo is available at:

https://github.com/AntoineSoetewey/datanalyze

Thanks in advance!

### Theme version

v2.5.2

### Hugo version

v0.101.0

Chrome, Safari

### Relevant Hugo log output

_No response_

Jh123x:
- [x] fix: close menu on click

Closes #460

### DIFF
--- a/assets/js/menu.js
+++ b/assets/js/menu.js
@@ -1,0 +1,6 @@
+/*
+    Closes the hamburger menu when a link is clicked.
+*/
+function close_menu() {
+    document.getElementById('menu-controller').checked=false
+}
--- a/exampleSite/config/_default/params.toml
+++ b/exampleSite/config/_default/params.toml
@@ -17,7 +17,7 @@ enableImageLazyLoading = true
 fingerprintAlgorithm = "sha256"
 
 [header]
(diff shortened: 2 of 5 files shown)


jpanther:
While this does solve the original issue, it introduces a new one in that this will violate content security policies that don't allow unsafe inline scripts. The way this needs to be implemented is by including this in an external script and then referencing the function from there.

jpanther:
Thanks, that change addresses the inline issue. I think we should also duplicate this into the `hybrid` header too otherwise the behaviour won't be consistent between header layouts.

jpanther:
Thanks for the update, I'll merge this one in later today.

PR #714, opened 2023-11-21
Remove unnecessary execute permissions

sorairolake:
Markdown and image files are not executable file, so I removed the execute permission from them. `release.sh` may be executable file, so I maintained the execute permission for this.


jpanther:
Thanks for the PR, @sorairolake. I'm happy to accept this change and will merge it in now.

PR #719, opened 2023-11-25
feat: add share options

Jh123x:
Changelog:
1. Add telegram sharing option
2. Add line sharing option


jpanther:
I'm happy to add this functionality however this PR needs a couple of changes:

- [x] Change the Line SVG to use the [icon from FontAwesome](https://fontawesome.com/icons/line?f=brands&s=solid) (and modify the asset to match the existing SVGs in the theme in terms of structure and adding `fill="currentColor"` to the `path` so it is themed correctly)
- [x] Add the required i18n strings for `sharing.telegram` and `sharing.line` (including placeholders in all the languages other than English, as required)

jpanther:
Wow, you really went above and beyond with the i18n strings. I was only expecting blank commented out placeholders but you've managed to actually translate them all! I'm happy to merge this one in. I'll get to it later today.

PR #723, opened 2023-11-26
remove screenshot shortcode

stereobooster:
But before merging checkout `/congo/docs/hosting-deployment/` page. Some screenshots look "broken".

I think the main problem here is resizing in Hugo. Maybe:

- instead of `2x` use specific flag for screenshots, which would do two things:
  - it will not resize image (serve as is)
  - it will divide sizes by two
- Introduce one more flag `original=true`
  - it will not resize image (serve as is)


jpanther:
The stale bot is putting me to shame. I'm still keen to work towards removing this shortcode but I haven't had any time to look into it!

stereobooster:
those soulless machine. Hey, bot, have some patience

PR #725, opened 2023-11-27
refactor: refactor headers to use same code

### LINKED ISSUE #721: Remove code repetition in header layouts
A long standing todo list item for this project is to refactor the header layouts to reduce the code repetition between the various layouts. Currently the `hybrid` layout duplicates the code from `basic` and `hamburger` and leads to issues when they are not kept in parity.

There's also potential to extract parts of the logic into functions so they can be reused in a more generic way. This would then allow more header layouts to be created without the overhead of maintaining support for existing features in every layout.

I'm creating this as an issue in case there's anyone in the community who has the time to look at submitting a PR for this change.

Jh123x:
- [x] refactor: refactor contents to its own directories
- [x] refactor: share `hybrid/basic/hamburger` content
- [x] fix: fix mobile hybrid

Closes #721


jpanther:
Sorry for leaving this PR around so long but one of the things I've been contemplating with this is whether this change is too simple.

For instance, I look at both the hamburger and basic content files and there's still a lot of repetition between the two. One of the goals I have in mind for this issue is to extract a whole bunch of the repetition like the appearance switcher and search into separate partials so that it's super easy to reuse them across a range of header templates.

This would also make it even easier for people to be able to override just one component of the header without having to make customisations to the entire block.

Jh123x:
> Sorry for leaving this PR around so long but one of the things I've been contemplating with this is whether this change is too simple.
> 
> For instance, I look at both the hamburger and basic content files and there's still a lot of repetition between the two. One of the goals I have in mind for this issue is to extract a whole bunch of the repetition like the appearance switcher and search into separate partials so that it's super easy to reuse them across a range of header templates.
> 
> This would also make it even easier for people to be able to override just one component of the header without having to make customisations to the entire block.

Will continue again based on the current dev branch once I have more time 👍

PR #726, opened 2023-11-27
fix: add image fix

### LINKED ISSUE #600: Images flipped 90 degrees

Jh123x:
- [x] fix: fix img flip
- [ ] test: test with image (No image to test)

Please let me know if anyone has an image to test

Closes #600

### DIFF
--- a/layouts/partials/picture.html
+++ b/layouts/partials/picture.html
@@ -81,6 +81,12 @@
         type="image/webp" 
       />
     {{ end }}
+    {{ with $img }}
+      {{ $orientation := 1 }}
+      {{ with .Exif }}
+        {{ $orientation = .Tags.Orientation }}
+      {{ end }}
+      {{ if eq $orientation 1 }}
       <img
         src="{{ .RelPermalink }}"
         width="{{ $width }}"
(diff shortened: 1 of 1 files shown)


stereobooster:
I think, it would be nice to have at least one such image in the repo to verify it's fixed (and verify in the future that it is not broken by any consequent change)

layouts/partials/picture.html
```diff
@@ -81,6 +81,12 @@
         type="image/webp" 
       />
     {{ end }}
+    {{ with $img }}
```
stereobooster: there is `{{ with $img }}` on top, so this one is redundant

stereobooster:
You can do something like this, to reduce repetition

```hugo
    {{ $s := newScratch }}
    {{ $s.SetInMap "sizes" "330w" 330 }}
    {{ $s.SetInMap "sizes" "660w" 660 }}
    {{ if gt .Width 1024 }}
      {{ $s.SetInMap "sizes" "1024w" 1024 }}
    {{ end }}
    {{ if gt .Width 1320 }}
      {{ $s.SetInMap "sizes" "2x" 1320 }}
    {{ end }}
    {{ if le .Width 1320 }}
      {{ $s.SetInMap "sizes" (printf "%dw" .Width) .Width }}
    {{ end }}

...

          srcset="
            {{ range $key, $size := $s.Get "sizes" }}
              {{ ($img.Resize (printf "%dx webp %s" $size $rotation)).RelPermalink }} {{ $key }},
            {{ end }}"

          srcset="
            {{ range $key, $size := $s.Get "sizes" }}
              {{ ($img.Resize (printf "%dx %s" $size  $rotation)).RelPermalink }} {{ $key }},
            {{ end }}"
```

to reduce repetition

Jh123x:
> You can do something like this, to reduce repetition
> 
> ```
>     {{ $s := newScratch }}
>     {{ $s.SetInMap "sizes" "330w" 330 }}
>     {{ $s.SetInMap "sizes" "660w" 660 }}
>     {{ if gt .Width 1024 }}
>       {{ $s.SetInMap "sizes" "1024w" 1024 }}
>     {{ end }}
>     {{ if gt .Width 1320 }}
>       {{ $s.SetInMap "sizes" "2x" 1320 }}
>     {{ end }}
>     {{ if le .Width 1320 }}
>       {{ $s.SetInMap "sizes" (printf "%dw" .Width) .Width }}
>     {{ end }}
> 
> ...
> 
>           srcset="
>             {{ range $key, $size := $s.Get "sizes" }}
>               {{ ($img.Resize (printf "%dx webp %s" $size $rotation)).RelPermalink }} {{ $key }},
>             {{ end }}"
> 
>           srcset="
>             {{ range $key, $size := $s.Get "sizes" }}
>               {{ ($img.Resize (printf "%dx %s" $size  $rotation)).RelPermalink }} {{ $key }},
>             {{ end }}"
> ```
> 
> to reduce repetition

Thanks for the suggestion. I am still new to hugo and have to read up on how this works first before implementing it ><

jpanther:
I think an issue with this implementation is that it's only dealing with one orientation. Ideally this needs to handle all the possible values and rotate the image accordingly.

PR #730, opened 2023-11-29
fix bug in picture with params

stereobooster:
I found a bug in implementation. Apparently `$.Page.Site.Params...` always evaluates to undefined in the context of `picture.html`. One way to fix it is always pass params explicitly. Other way would be to pass `.Page`, I guess

### DIFF
--- a/layouts/_default/_markup/render-image.html
+++ b/layouts/_default/_markup/render-image.html
@@ -20,7 +20,9 @@
 
 <figure>
   {{- with $img -}}
-    {{ partial "picture.html" (dict "img" . "alt" $altText "class" $class "x2" $x2) }}
+    {{ $lazy := $.Page.Site.Params.enableImageLazyLoading | default true }}
+    {{ $webp := $.Page.Site.Params.enableImageWebp | default true }}
+    {{ partial "picture.html" (dict "img" . "alt" $altText "class" $class "x2" $x2 "lazy" $lazy "webp" $webp) }}
   {{- else -}}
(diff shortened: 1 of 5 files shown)


wolfspyre (commented):
is it “better” to eval the current page’s params context as well for $lazy?
 thus enabling it to be configurable per page? 

i honestly dunno.. just asking

stereobooster:
I don't quite understand what you propose. Can you post a code snippet?

jpanther:
Does this need to reference `.Page` anyway? Given these are site config variables, can't we just access the global `.Site` directly?

wolfspyre:
> Does this need to reference `.Page` anyway? Given these are site config variables, can't we just access the global `.Site` directly?

wouldn’t using .Page enable one to disable lazy (or configure other attrs) at a per-page level? 

if the only knobs this looks at are site-scoped doesn’t one lose this? 

When asking on the hugo forums on how to enable a clean lookup hierarchy for partial config variables  in site config[^] , I was steered towards Page scoped variables and using cascade… 

not sure if that’s helpful or not…  but it steers my thinking towards setting things to look for page-scoped knobs and set the page vars’ default to site … 


[^1]:  Ie: how can a partial lookup .Site.Params.$(myscope,or-maybe-fallback-to-Type-here).sprocket.gizmo.enabled

jpanther:
Well potentially, but why would you need to adjust the lazy loading preference on a per-page basis? That seems like something that you'd only control at a site level.

PR #751, opened 2023-12-17
:lipstick: update CSS Schemas to include TW3.3+ 950 color variants

wolfspyre:
expand the CSS profiles to include the 950 variants of colors added in TailwindCSS 3.3:

https://tailwindcss.com/blog/tailwindcss-v3-3#extended-color-palette-for-darker-darks

### DIFF
--- a/assets/css/schemes/avocado.css
+++ b/assets/css/schemes/avocado.css
@@ -2,39 +2,39 @@
 :root {
   --color-neutral: 255, 255, 255;
   /* Stone */
-   --color-neutral-50: 250, 250, 249;
+  --color-neutral-50: 250, 250, 249;
   --color-neutral-100: 245, 245, 244;
   --color-neutral-200: 231, 229, 228;
   --color-neutral-300: 214, 211, 209;
   --color-neutral-400: 168, 162, 158;
   --color-neutral-500: 120, 113, 108;
-  --color-neutral-600:  87,  83,  78;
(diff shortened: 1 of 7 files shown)


jpanther:
Happy to add these in, although I'm not so sure about some of this whitespace padding. The space before the '50' variant is particularly obtuse. I'm surprised Prettier isn't overriding that?

wolfspyre:
an here I thot I was helping there, by aligning the values such that regardless of the value, or the previous values, they’d all line up when eyeballing.. 
this is your baby, I’m not tryin to be a pain. 

is yer preference that the hyphens align vs the colons?

That’s simple enough. LMK and I’ll push a fix

jpanther:
Thanks for updating the formatting, I'll merge this in.

wolfspyre:
> Thanks for updating the formatting, I'll merge this in.

ABSOLUTELY. My **INTENT** is to be **HELPFUL**. I jus occasionally need a bit more guidance than one might assume.

**embarassed look** I'd never used prettier before.. I'd jus sorta assumed it was being managed in CI.. 

having looked at it now a bit, I'm not sure how I feel about it... I haz opinions about some of the things it does to one's code layout...  but ¯\\\_(ツ)_/¯   I'll deal with that when I have practical examples to articulate my opinion.

...One of the "*joys*" of having a bit of a weird brain ;)

PR #756, opened 2023-12-21
✨ more profiles and share links

Jh123x:
-  [x] feat: Share to Weibo
-  [x] chore: add to all i18n weibo config
-  [x] feat: Mendeley Profile

image


exampleSite/config/_default/languages.en.toml
```diff
@@ -22,4 +22,5 @@ copyright = "&copy; 2023 Congo contributors"
     { facebook = "https://facebook.com/" },
     { linkedin = "https://linkedin.com/" },
     { youtube = "https://youtube.com/" },
+    { mendeley = "https://www.mendeley.com/" },
```
jpanther: As per the other PR, I prefer to not push these types of changes to the example site.

PR #762, opened 2023-12-27
:memo: Adds personal site to users example

jvanderaa:
Adds personal page to the user examples. 

Thank you for the effort that you put in!


jpanther:
Thanks for using Congo, @jvanderaa! I'll merge this change in and your site will appear on the public list when the next theme release is published.

PR #764, opened 2023-12-30
fix: fix extra ß symbol

Jh123x:
- [x] fix: remove additional symbol


jpanther:
Thanks for picking this up. Merging it in now.

PR #766, opened 2024-01-04
remove unneccessary ß in single.html

cambid:
This PR removes an unnecessary ß in the single.html introduced in the latest upgrade of tailwind.


jpanther:
Duplicate of #764

PR #767, opened 2024-01-05
Add site to list of users

rshmhrj:


jpanther:
Thanks for adding your site to the list, @rshmhrj! I'll merge this change in and it will appear on the public list when the next theme release is published.

PR #769, opened 2024-01-06
Add jamesjarvis.io to list of users

jamesjarvis:
Thanks for the amazing theme!
I rewrote my site a while ago to get rid of GatsbyJS, and this theme was an amazing replacement, with a similar design + easy configuration with Hugo!
I wrote a little bit about the benefits of it here: https://jamesjarvis.io/posts/2023/07/blogging-site/

Again, thanks for this!


jpanther:
Thanks for using Congo, @jamesjarvis! I'll merge this change in and your site will appear on the public list when the next theme release is published.

PR #777, opened 2024-01-12
Add jnsgr.uk to list of users

jnsgruk:
Simple change to add my personal site to the list of sites using `congo`. Also fixes a very minor alignment issue in one of the other entries as a drive-by.


jpanther:
Thanks for using Congo, @jnsgruk! I'll merge this change in and your site will appear on the public list when the next theme release is published.

PR #778, opened 2024-01-13
feat: use meta tag to redirect externalURLs

Jh123x:

### DIFF
--- a/layouts/partials/head.html
+++ b/layouts/partials/head.html
@@ -120,6 +120,10 @@
   {{ with .Site.Params.verification.yandex }}
     <meta name="yandex-verification" content="{{ . }}" />
   {{ end }}
+  {{ with $.Params.externalUrl }}
+    <meta http-equiv="refresh" content="0; url={{ . }}" />
+  {{ else }}
+  {{ end }}
   {{/* Social */}}
   {{ template "_internal/opengraph.html" . }}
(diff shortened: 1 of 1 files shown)


jpanther:
Thanks for the PR, but why is this required? Links to external URLs already send you to the correct destination.

Jh123x:
> Thanks for the PR, but why is this required? Links to external URLs already send you to the correct destination.

In cases where the users go to the URL directly, the webpage itself does not redirect to the external URL.

jpanther:
Interesting. I don't see this issue as I always set those pages to not be built to prevent that from happening. I don't think it's good for SEO to have pages that are just redirects. It's much better to be linking to the original content and forming backlinks. Is there a particular reason you still generate output for external URL pages?

Jh123x:
The use case for me is to redirect links which have been migrated to another website. Should this be under another property instead?

jpanther:
Hmm okay, no that use case is valid and one that I hadn't considered. I don't see any reason why the same parameter couldn't serve both purposes. If you don't need the leaf pages you can tell Hugo not to generate them, but in your use case where you want them, then you can leave them on and have them redirect.

I'll accept the PR, although is there a reason you coded this with an `{{ else }}` block? It seems redundant.

jpanther:
This is tested and working so I'll merge it in.

PR #785, opened 2024-01-19
Fix picture partial img/source attributes

pomeloy:
Remove duplicate src attribute from image tag.
Remove not allowed src attribute from source tag.
Add default sizes attributes to img and source tag.
Fix inconsistent srcset width.


jpanther:
Thanks for picking up these issues. It's great to fix them before the new `picture` partial is released. I'm happy to merge them in.

PR #786, opened 2024-01-19
Do not render article-meta figure if there is nothing to display

pomeloy:


jpanther:
Thanks for the PR. I'm happy to accept this change.

PR #789, opened 2024-01-19
💄 Fix Safari status bar style

jukremer:
Add status bar style meta tag in head.html, add setting the same value in appearance.js


jpanther:
Thanks, I didn't even know this tag existed. Happy to merge it in.

jukremer:
Glad I could help.

PR #791, opened 2024-01-20
Revert "💄 Fix Safari status bar style"

pomeloy:
Reverts jpanther/congo#789

The theme-color tag is [supported](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/meta/name/theme-color#browser_compatibility) by all Apple devices released since 2015 (iOS 15, available from 2021).

In addition, the meta tag apple-mobile-web-app-capable has no effect unless full-screen mode as described in apple-[apple-mobile-web-app-capable](https://developer.apple.com/library/archive/documentation/AppleApplications/Reference/SafariHTMLRef/Articles/MetaTags.html#//apple_ref/doc/uid/TP40008193-SW3) is specified. A regular congo webpage should not run in full-screen mode.

### DIFF
--- a/assets/js/appearance.js
+++ b/assets/js/appearance.js
@@ -8,13 +8,9 @@ function getCSSValue(varName) {
 
 function setThemeColor() {
   var metaThemeColor = document.querySelector("meta[name=theme-color]");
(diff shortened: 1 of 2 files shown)


jpanther:
Having looked into this more closely now, I tend to agree. I did test the PR that introduced this change yesterday and couldn't see any negative consequences but it seems pointless if `theme-color` is already covering this use case.

PR #806, opened 2024-02-06
feat: add js to check height and decide if to top is required

### LINKED ISSUE #772: showScrollToTop and minimum 200 characters
### Issue description

Hi, thanks for a great theme! And great documentation, it was easy to get started with such thorough docs 🙇 

To get started, I just created a dummy post with some random things, including some parts of the [markdown.md](https://github.com/jpanther/congo/blob/dev/exampleSite/content/samples/markdown/index.md) document. The contents of the file are provided below.

As you can see, it is a pretty long (tall?) document, so I was expecting that the `scrollToTop` button would appear for this article. But no, I could not get it working and had a hard time debug why, in particular since this was my only long document.

In the end, I figured out why, and it is due to this clause of minimum 200 words required: https://github.com/jpanther/congo/blob/c8f648d5c2f060592461b1ac41fe9a764422e7ee/layouts/_default/baseof.html#L32

I suggest to either:

1. Use some other logic for the minimum size of document to enable the `scrollToTop` button, as it seems number of words is to coarse. The document below might not have many words, but it is definitely long enough to benefit from that button IMO.
2. Reduce the minimum number of words. I am not sure how much it would "hurt" if this button gets shown for shorter articles?
3. Document [here](https://jpanther.github.io/congo/docs/configuration/#theme-parameter) under `footer.showScrollToTop` that there is a minimum requirement for `x` words for the button to be shown in an article. That would have helped me a lot in this debugging :) 

**Start of my dummy document:**

---
I am regular text. This is okay.

## Heading H2

Here is some Go for ya.

```go
package main

func main() {
    fmt.Println("Hello, World!")
}
```

## Pictures are nice

But can I add them via reference tags? Nope, that doesn't work. Okay, I can use just as a convenience when writing, and then change to regular Markdown syntax on publishing.

UPDATE: all images needs to be moved into the `static` folder anyway, so some "post processing" before publishing would be required any way.

Alt text

A dog

## Tables

Tables aren't part of the core Markdown spec, but Hugo supports supports them out-of-the-box.

| Name  | Age |
| ----- | --- |
| Bob   | 27  |
| Alice | 23  |

### Inline Markdown within tables

| Italics   | Bold     | Code   |
| --------- | -------- | ------ |
| _italics_ | **bold** | `code` |

### Ordered List

1. First item
4. Second item
5. Third item

### Unordered List

- List item
- Another item
- And another item

#### Nested list

- Fruit
  - Apple
  - Orange
  - Banana
- Dairy
  - Milk
  - Cheese

## Other Elements — abbr, sub, sup, kbd, mark

<abbr title="Graphics Interchange Format">GIF</abbr> is a bitmap image format.

H<sub>2</sub>O

X<sup>n</sup> + Y<sup>n</sup> = Z<sup>n</sup>

Press <kbd>CTRL</kbd>+<kbd>ALT</kbd>+<kbd>Delete</kbd> to end the session.

Most <mark>salamanders</mark> are nocturnal, and hunt for insects, worms, and other small creatures.

---

**End dummy document.**

### Theme version

v2.7.6

### Hugo version

hugo v0.121.1-00b46fed8e47f7bb0a85d7cfc2d9f1356379b740+extended darwin/arm64

WebKit (Safari)

### URL to sample repository or website

_No response_

### Hugo output or build error messages

_No response_

Jh123x:
Set scroll to top element to available when page height -150 > browser height.
`-150` is based on testing

Closes #772


jpanther:
I'm not against making this a JS feature as frankly the current implementation using only CSS is a bit hacky, but I still think it should honour the `.Site.Params.footer.showScrollToTop` parameter.

Jh123x:
> I'm not against making this a JS feature as frankly the current implementation using only CSS is a bit hacky, but I still think it should honour the `.Site.Params.footer.showScrollToTop` parameter.

Hmm, let me try it in another way

Jh123x:
It should also obey the rule in the config now

jpanther:
Thanks, I'll merge this in now!

PR #821, opened 2024-03-03
Add techwolf12.nl as user of the theme

Techwolf12:
Thanks for the great theme!


jpanther:
Thanks for using Congo, @Techwolf12! I'll merge this change in and your site will appear on the public list when the next theme release is published.

PR #833, opened 2024-03-18
add kylecapehart.com as user of the theme

k-capehart:
- add kylecapehart.com as user of the theme

Thanks for the awesome theme! I love it.


jpanther:
Thanks for using Congo, @k-capehart! I'll merge this change in and your site will appear on the public list when the next theme release is published.

PR #840, opened 2024-03-26
Fixing Incorrect Language Codes in baseof.html Template

TheSethRose:
This change fixes incorrect language codes referenced in the `baseof.html` template by updating the lang attribute to use `.Site.Language.LanguageCode` instead of `site.Language.LanguageCode` and `.Site.Language.LanguageDirection` instead of `site.Language.LanguageDirection`, ensuring the HTML language matches the configuration.

### Current Code

```html
<html
  lang="{{- site.Language.LanguageCode | default "en" -}}"
  dir="{{- site.Language.LanguageDirection | default "ltr" -}}"
  class="scroll-smooth"
  data-default-appearance="{{- .Site.Params.defaultAppearance | default "light" -}}"
  data-auto-appearance="{{- .Site.Params.autoSwitchAppearance | default "true" -}}"
>
```

### Updated Code

```html
<html
  lang="{{- .Site.Language.Lang | default "en" -}}"
  dir="{{- .Site.Language.LanguageDirection | default "ltr" -}}"
  class="scroll-smooth"  
  data-default-appearance="{{- .Site.Params.defaultAppearance | default "light" -}}"
  data-auto-appearance="{{- .Site.Params.autoSwitchAppearance | default "true" -}}"
>
```

The updated code uses the correct properties to set the language and direction attributes. This will make the HTML output align with the site's language configuration settings.

### DIFF
--- a/layouts/_default/baseof.html
+++ b/layouts/_default/baseof.html
@@ -2,8 +2,8 @@
 {{- partial "partials/functions/init.html" . -}}
 <!doctype html>
 <html
-  lang="{{- site.Language.LanguageCode | default "" -}}"
-  dir="{{- site.Language.LanguageDirection | default "ltr" -}}"
+  lang="{{- .Site.Language.LanguageCode | default "" -}}"
+  dir="{{- .Site.Language.LanguageDirection | default "ltr" -}}"
   class="scroll-smooth"
   data-default-appearance="{{- .Site.Params.defaultAppearance | default "light" -}}"
   data-auto-appearance="{{- .Site.Params.autoSwitchAppearance | default "true" -}}"


jpanther:
Thanks for the PR, I'm just trying to get an understanding of what issue this is solving? Is there a particular use case where the current `site` function is not working correctly?

wolfspyre:
https://discourse.gohugo.io/t/site-vs-site-vs-site/49252/4

~1w old thread on the forum says the difference between Site and site is mostly multilang related; so this may be legit… idk, haven’t poked at it, but smells worth exploring to me

jpanther:
Part of the reason I query it though is because it used to use `.Site` and was changed to use `site` when the multilang support was added to the theme and so I can only suppose that it was required to solve some issue at the time. I can't see this not working in its current implementation but I don't use the multilang myself so it could be an edge case I'm not seeing.

TheSethRose:
Honestly, I am not sure that I remember the specific issue I was trying to solve with this PR. I was working with this code and found that it was broken for me. Making the adjustment to use .Site instead of site got it working in my case. However, I ended up scrapping the changes and going a different route with my project. I apologize that I cannot provide additional context on the specific case I ran into.

You raise a good point and If the current implementation using site is working well, especially for multilingual scenarios, then it may be best to stick with that.

Since I'm no longer pursuing this change in my own project, feel free to close this PR if you don't think it's necessary or beneficial to switch back to using .Site at this time. I defer to your judgment as the theme author and maintainer.

Sorry I couldn't provide more details on the original issue I was trying to solve. Let me know if you have any other questions!

jpanther:
Thanks for the extra comments. I'll close this for now as I can't see any issues arising from the current implementation. Happy to reopen if anything arises that suggests otherwise.

PR #844, opened 2024-03-28
Add hosni.info as user of the theme

hosni:


jpanther:
Thanks for using Congo, @hosni! I'll merge this change in and your site will appear on the public list when the next theme release is published.

PR #849, opened 2024-04-01
:sparkles: add defaultThemeColor option

ItsSunnyMonster:
Some applications (such as Discord) use the `theme-color` meta tag to display an accent color on its embeds. Currently the theme-color will be changed when the page loads by a script to reflect the current theme of the page (dark or light), but Discord ignores this and takes the value off of the original HTML. Therefore we can safely set this tag to some color specified by the user and not break the theme-aware theme-color functionality.

P.S. I don't speak Japanese so it will be great if anyone could translate the documentation for the new param.

### DIFF
--- a/config/_default/params.toml
+++ b/config/_default/params.toml
@@ -9,6 +9,8 @@ colorScheme = "congo"
 defaultAppearance = "light" # valid options: light or dark
 autoSwitchAppearance = true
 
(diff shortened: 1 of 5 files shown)


jpanther:
Thanks for the PR, @ItsSunnyMonster. I'm happy to add this feature in, and the documentation you've supplied is super helpful! Will merge it in now.

PR #850, opened 2024-04-02
add personal site to users list

mattstrayer:
https://www.mattstrayer.com


jpanther:
Thanks for using Congo, @mattstrayer! I'll merge this change in and your site will appear on the public list when the next theme release is published.

PR #859, opened 2024-04-16
Add noamlerner.com to users

bugok:
Adding my personal English / Hebrew blog to the list of users.


jpanther:
Thanks for using Congo, @bugok! I'll merge this change in and your site will appear on the public list when the next theme release is published.

PR #871, opened 2024-05-04
fix: rename author to params.author

jneidel:
Addresses this warning:

```
WARN  [CONGO] Theme parameter `author` block in `languages.xx.toml` has been renamed to `params.author`. Please update your site configuration.
```


jpanther:
The warning quoted is provided by the theme and has nothing to do with the change submitted in this PR. I don't believe this is a valid change. Please file a bug report if you believe you have found a bug related to this warning.

PR #872, opened 2024-05-05
Add Norwegian localisation

torbbang:
Added support for Norwegian Bokmål(Norsk Bokmål).


jpanther:
Thanks for this translation, @torbbang! I'll merge this in now.

PR #875, opened 2024-05-09
Fix search to always fetch index.json from /

jneidel:
I got 404s when deploying to netlify because baseURL was a different domain from where it was deployed.

Sure, you could say that I should fix the config on my end (and I will.)

But this could still be simplified to just always fetch from the site root. Specifying the domain is wholly unnecessary.


jneidel:
Nevermind, it is needed for multiple languages.

PR #878, opened 2024-05-12
Add jneidel.com as user

jneidel:
Also remove another user that throws a 404


jpanther:
Thanks for using Congo, @jneidel! I'll merge this change in and your site will appear on the public list when the next theme release is published.

jneidel:
I really like the look and feel of it and it offers all of the functionality I was looking for.
I can see that it was really well thought out.
Thank you for creating it!

PR #887, opened 2024-05-24
Add clementfouque.com site to the list of users

clement-fouque:


jpanther:
Thanks for using Congo, @clement-fouque! I'll merge this change in and your site will appear on the public list when the next theme release is published.

PR #896, opened 2024-06-12
add victoryuan.com to users

wvictor14:
Thank you for creating this nice theme!


jpanther:
Thanks for using Congo, @wvictor14! I'll merge this change in and your site will appear on the public list when the next theme release is published.

PR #902, opened 2024-06-20
Add chenyu.blog as user

BinaryFiddler:
Adding chenyu.blog to the list of users of congo


jpanther:
Thanks for using Congo, @BinaryFiddler! I'll merge this change in and your site will appear on the public list when the next theme release is published.

PR #921, opened 2024-08-25
Replace .Site.IsMultiLingual with hugo.IsMultilingual

jneidel:
Addresses this deprecation warning:

```
WARN  deprecated: .Site.IsMultiLingual was deprecated in Hugo v0.124.0 and will be removed in a future release. Use hugo.IsMultilingual instead.
```

[hugo.IsMultilingual](https://gohugo.io/functions/hugo/ismultilingual/) was introduced in v0.124.0.
So merging this PR would effectively make that the minimum required version. (See failing pipelines on initial commit, which are only using v0.119.)
Maybe you want to hold off on merging this for the time being :smile:

### DIFF
--- a/layouts/partials/footer.html
+++ b/layouts/partials/footer.html
@@ -4,7 +4,7 @@
     <nav class="pb-4 text-base font-medium text-neutral-500 dark:text-neutral-400">
       <ul class="flex list-none flex-col sm:flex-row">
         {{ range .Site.Menus.footer }}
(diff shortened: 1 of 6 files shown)


jpanther:
Thanks for the PR. I've held off on merging this for a little while as I don't want to force people to have to bump their Hugo versions unnecessarily but it's been out for 6 months now which I think is a reasonable time for people to update their dependencies. I'll merge this in and update the theme minimum Hugo version to v0.124.0.

PR #932, opened 2024-09-06
📝 Add socrabytes.github.io to the user list

socrabytes:
This commit adds my personal site to the user list in the `users` content to showcase users who are leveraging the Congo theme.


jpanther:
Thanks for using Congo, @socrabytes! I'll merge this change in and your site will appear on the public list when the next theme release is published.

PR #933, opened 2024-09-07
feat: Add sharing link and icons for XING

Zeruxky:
This PR adds sharing links and icons for the business social media platform [XING](https://xing.com). I added translations of the sharing text for German and English. Other i18n files contains a placeholder that is commented out.

Changelog:

- Add icon for XING to be displayed within template and at the example page.
- Add option to add link to own XING profile.
- Add option to share a post on XING.
- Add text in German and English when the user hovers above the share button.

PR #948, opened 2024-10-07
fix: Removes follow_url part of sharing link for XING

Zeruxky:
This PR removes the follow_url part of the sharing link for XING, which contained the URL to my personal XING profile. I accidentally forgot to remove this in my previous PR (#933). The sharing functionality should be unaffected by this PR.

Sorry for the inconveniences!


jpanther:
Thanks, I totally missed that one too. I'll merge it in now.

PR #952, opened 2024-10-15
Update index.md

pitermarx:
Add user

I have a large history of posts, and the search feature is a bit slow for me.
If you could take a look I would appreciate it!


pitermarx:
I did this to minimize the effect on my site
https://github.com/pitermarx/pitermarx.github.io/commit/4d62285b8cedd13f873357daa0b52170f8dc954a

jpanther:
Thanks for using Congo, @pitermarx! I'll merge this change in and your site will appear on the public list when the next theme release is published.

I do like your loading state for the search function too. I'd be happy to look at a PR for that to be included as an enhancement if you'd like to submit it.

PR #955, opened 2024-10-20
🩹 Add fill="currentColor" to Bluesky icon

gaganpreet:
fill="currentColor" is missing in the Bluesky icon making it stand out from other icons (see image). This was the only icon with the missing fill attribute and I presume it was unintentionally missed.

Screenshot_20241020_124953


joaroyen:
I would also like this pull request merged and released

dotnwat:
+1 to this. I hope this goes in

jpanther:
Thanks for picking this up. Merging in now.

PR #961, opened 2024-11-19
📝 docs: add reference to hugo.toml as a configuration file

hugo-sid:
Since January 17, 2023, starting with [v0.110.0](https://github.com/gohugoio/hugo/releases/tag/v0.110.0), `hugo.toml` is the default configuration file


jpanther:
Thanks for picking up on that change. I'll merge this in.

PR #975, opened 2024-12-06
Update package-lock versions to 2.10.0

jneidel:
Got this git diff after pulling 2.10.0 and running npm install:

```diff
 {
   "name": "hugo-congo-theme",
-  "version": "2.9.0",
+  "version": "2.10.0",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "hugo-congo-theme",
-      "version": "2.9.0",
+      "version": "2.10.0",
       "hasInstallScript": true,
       "license": "MIT",
       "dependencies": {
```

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "hugo-congo-theme",
-  "version": "2.9.0",
+  "version": "2.10.0",
   "lockfileVersion": 2,
   "requires": true,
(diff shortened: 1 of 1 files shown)


jpanther:
Thanks, I'll fix this in future so the script that builds releases updates the lock file so the correct version is pulled across.

PR #984, opened 2025-01-01
assets: add rss icon

sudomateo:
This patch adds the Font Awesome `rss` icon which I believe falls under the free tier.

The icon is sourced from https://fontawesome.com/icons/rss.

This allows user to create an RSS link in their socials or menu bar.


sudomateo:
I didn't realize that deleting the fork would close the pull request. Oh well, someone else can submit a new pull request. I don't like keeping forks around for repositories I infrequently touch.

PR #985, opened 2025-01-01
Docs: List languages.en.toml in config

hugovk:
This file is also shown at https://github.com/jpanther/congo/tree/stable/config/_default

And needed in subsequent setup steps:

https://jpanther.github.io/congo/docs/getting-started/


jpanther:
Thanks for this change, I'm happy to merge it in.

jpanther (approved):

PR #986, opened 2025-01-02
🐛 Fix: Correct absolute path for shebang in shell script

aminelch:
**Ensure the shebang uses the absolute path to the interpreter.**
The interpreter should always be specified by absolute path to ensure that the script can be executed from any directory. When it's not, it's generally a typo.

**Problematic code:**
```bash
#!bin/sh
echo "Hello World"
```
**Correct code:**
```bash
#!/bin/sh
echo "Hello World"
```
See [https://github.com/koalaman/shellcheck/wiki/SC2239](https://github.com/koalaman/shellcheck/wiki/SC2239)


jpanther:
Thanks for this change. I'm happy to merge it in.

jpanther (approved):

aminelch:
> Thanks for this change. I'm happy to merge it in.

welcome 🙏

PR #990, opened 2025-01-03
:bug: Fix: incorrect translation link in multihost mode #988

### LINKED ISSUE #988: Incorrect translation link in multihost mode
I found that maybe a bug(?)  if the hugo site is configured in multihost mode(different languages use different base urls).

I checked translations.html's code and found that different language's url is using RelPermalink as href target.This might have some problems if the site is in multihost mode.

So, I think would be more suitable using Permalink in multihost mode. Please correct me if this is wrong. I will make a PR later.

shane-yao:
Use permalink for multihost mode and leave relpermalink unchanged in non multihost mode.

Since multihost mode's urls aren't correct currently. This will minimal the impact of other users.

### DIFF
--- a/layouts/partials/translations.html
+++ b/layouts/partials/translations.html
@@ -35,7 +35,7 @@
                 <div class="flex flex-grow">
                   {{ if eq $pageLang .Lang }}
                     <a
-                      href="{{ $translation.RelPermalink }}"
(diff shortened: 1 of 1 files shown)


jpanther:
Thanks for this fix. I'm happy to merge it in.

jpanther (approved):

PR #994, opened 2025-01-16
📝 Modify noamlerner.com's base url in the users page

bugok:
I originally added my website in https://github.com/jpanther/congo/pull/859. Since then, I decided to update the base url, and I'll soon remove the www domain, so I'd like to update to the new domain, which is already active.


jpanther:
Thanks for keeping the URL up-to-date, I'll merge this in.

PR #996, opened 2025-01-22
Add euangelos.linardos.gr to the user list

eualin:
Does what it says on the tin.


jpanther (approved):

jpanther:
Thanks for using Congo, @eualin! I'll merge this change in and your site will appear on the public list when the next theme release is published.

PR #997, opened 2025-01-23
🐛 picture partial: use absolute permalinks

angristan:
To fix image rendering in feed readers, in case you have images in the content of the RSS feed.


jpanther:
Thanks for the PR, I'm happy to adopt this change but will do it outside this PR as the project has been restructured to match upstream Hugo changes.

PR #1000, opened 2025-01-27
:memo: Add joshquinlan.co.uk to user list

ghost:
Recently started using the Congo theme for my site, and would love to be added to the list of Users.


jpanther:
Thanks for using Congo, @DonQuinleone! I'll merge this change in and your site will appear on the public list when the next theme release is published.

PR #1002, opened 2025-02-06
Add blog.sky123.top to user list

sky-max-hub:
I really love the Congo theme as the foundation for my Hugo blog site and have made many modifications to it. I would be honored to be included in the list of Users.


jpanther:
Thanks for using Congo, @sky-max-hub! I'll merge this change in and your site will appear on the public list when the next theme release is published.

PR #1006, opened 2025-02-06
Rename config.toml to hugo.toml

deining:
This PR renames `config.toml` to `hugo.toml`, as suggested [here](https://gohugo.io/getting-started/configuration/#configuration-file).


jpanther:
Thanks for the PR. I wasn't aware that this had changed, but since this requires a lot of documentation adjustments and other minor changes that prevent it from being able to be merged, I'll work through this outside this PR and submit it separately.

PR #1007, opened 2025-02-06
Remove documentation on deprecated shortcode 'gist'

deining:
When running example site with latest released hugo version 0.143.1, a warning is displayed:

```
WARN  The "gist" shortcode was deprecated in v0.143.0 and will be removed in a future release.
See https://gohugo.io/shortcodes/gist for instructions to create a replacement.
```

This PR fixes that issue.


jpanther:
Thanks for this update. I'm going to reimplement 'gist' so it's bundled as a shortcode within the theme, but this documentation still needs to be removed, so I'll merge this change in.

PR #1008, opened 2025-02-09
Update index.md

brihadeesh:
added entry for peregrinator.site


jpanther:
Thanks for using Congo, @brihadeesh! I'll merge this change in and your site will appear on the public list when the next theme release is published.

PR #1010, opened 2025-02-18
add wendyliga.com

wendyliga:


jpanther:
Thanks for using Congo, @wendyliga! I'll merge this change in and your site will appear on the public list when the next theme release is published.

PR #1025, opened 2025-02-24
Added spinner for long searches

pitermarx:
Added a spinner for long searches
As seen in https://blog.pitermarx.com/
Animation


jpanther:
Thanks for this PR. I'm happy to merge it in!

PR #1028, opened 2025-03-02
fix: Don’t markdownify dates

### LINKED ISSUE #1027: date is shown incomplete using German localization
### Language

German (de-DE)

### What's the issue?

In German-speaking countries, a date like "2020-08-14" (the 14th of August 2020) is often written as "14. August 2020" (this is the localization of :date_long for de-DE). Due to the fact that the date in the article header, which shows the article date, is processed using the "markdownify" function, the date is interpreted as an ordered list starting at the Day (In the example, "14"):

```
<time datetime="2020-08-14 00:00:00 +0000 UTC">
  <ol start="14">
    <li>August 2020</li>
  </ol>
</time>
```

Since the ordered list has the style setting "list-style: none;", the Day (14) is not displayed at all.

This issue can be observed (as example) on the following pages: 
https://jpanther.github.io/congo/de/samples/icons/
https://jpanther.github.io/congo/de/samples/

### Theme version

v2.11.0

### Hugo version

v0.124.1

### Relevant Hugo log output

```shell

```

davidolrik:
To prevent “ordered list” markup in languages like Danish and German dates are no longer markdownifed.

Date formates like: `2. January 2006` would be turned into ordered lists because of the period after 2.

I didn't remove `emojify`, but I think it would also make sense to remove it as the output from the `dateFormat` probably wouldn't contain any emojis either.

Fixes #1027

### DIFF
--- a/layouts/partials/meta/date-updated.html
+++ b/layouts/partials/meta/date-updated.html
@@ -1,4 +1,4 @@
 <time datetime="{{ . }}">
-  {{- i18n "article.date_updated" (dict "Date" (partial "functions/date.html" .)) | markdownify | emojify -}}
+  {{- i18n "article.date_updated" (dict "Date" (partial "functions/date.html" .)) | emojify -}}
 </time>
 {{- /* Trim EOF */ -}}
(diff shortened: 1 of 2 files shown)


scorchio:
Thanks for the fix, I just ran into https://github.com/jpanther/congo/issues/1027 when trying to set up a usual Hungarian date formatting.

@jpanther could you please consider merging this? I think this fixes a fairly important bug for non-English sites, as for those currently, if you want to make sure a custom date format renders properly, you have to override this manually.

jpanther:
Thanks for this, sorry it's taken me such a long time to review it. Happy to merge it in!

PR #1032, opened 2025-03-06
Update index.md

lpaganelli:


jpanther:
Thanks for using Congo, @lpaganelli! I'll merge this change in and your site will appear on the public list when the next theme release is published.

PR #1038, opened 2025-03-21
add cmp1315.com to the list of users

AntoineSoetewey:
add cmp1315.com (a site for a medical center) to the list of users


jpanther:
Thanks again @AntoineSoetewey, I'll merge this one in.

PR #1054, opened 2025-04-23
attempt to fix theme from 0.146.x changes

wolfspyre:
Adapted the changes from https://github.com/nunocoracao/blowfish/pull/2055
( Credit goes to  [Ryan Gibson](https://github.com/ragibson) ) :) 
seems to build, at least.


wolfspyre:
builds successfully with 0.146.6 locally.. nasure if you wanna put gating logic for older versions.... 

I'd say this aught be a major version bump, as it likely causes breakage or unexpected behavior for adapted themes.... 

Not sure when hugo will consider themselves 1.x but... given the likelihood of other wobbles it feels cleaner (imo) to major rev to telegraph the significance... 

yer call tho. :) Congo's yer baby.

wolfspyre:
https://discourse.gohugo.io/t/conditional-overloading-of-an-embedded-shortcode-post-0-146-0/54481

The change to the figure shortcode in blowfish doesn't work. will update once I've clarity on the right way forward

wolfspyre:
Okay. Clarity recieved. 
https://github.com/gohugoio/hugo/issues/13553#issuecomment-2819173746
https://discourse.gohugo.io/t/conditional-overloading-of-an-embedded-shortcode-post-0-146-0/54481/3

TLDR:

The guidance forward is to create uniquely named wrappers.
The practice of reach-behind calling embedded shortcodes
   (and likely templates, but I was explicitly solely focusing on shortcodes in my questioning) 
is now officially and explicitly considered an antipattern to avoid. 

so we've got a pivot ahead, but at least a direction forward. 

I'll call it `congo-figure` unless you've a better idea?

wolfspyre:
mkay. patched, docs updated,  bumped to 3.0.0 to indicate the breakage. 
I'm sure you'll want to adjust stuff, but I wanted to at least get something that "works"

G-Snipes (approved):
This fix works for me when I spin up a local server.

ghost:
Any update on when this can be merged in @jpanther? This solution works well from my testing, and I'm personally keen to move off Hugo 0.145.

jpanther:
Thanks for the PR but I believe I've worked around this problem without needing to change the shortcode name. Apologies that it's taken me so long to get onto this - I've been off on other projects. I'll close this one for now but let me know if you spot any issues with my new changes.

PR #1055, opened 2025-04-26
Refined CS translation

Aidam7:
Refined the translations for CS, mostly just opiniated changes which sound better and are more commonly used than the current terms, except for word_count translation which was grammatically incorrect.

EDIT: After further inspection the current CS translation was most likely machine translated


jpanther:
Thanks for these updates, I'm happy to merge them in.

PR #1065, opened 2025-06-08
Add missing space in the LINE sharing link

rlan:
The URL spec for LINE is:

https://developers.line.biz/en/docs/messaging-api/using-line-url-scheme/#sending-text-messages

A space is missing between the URL and blog title. Without it, the blog URL and the first word of the blog title would be concatenated. The LINE messenger would then interpret that as a URL, which is an invalid link.

I applied the same fix in Blowfish. Check the deployment preview there for a working exampleSite:

https://github.com/nunocoracao/blowfish/pull/2213

### DIFF
--- a/data/sharing.json
+++ b/data/sharing.json
@@ -52,8 +52,8 @@
   "line": {
     "icon": "line",
     "title": "sharing.line",
-    "url": "https://line.me/R/share?text=%s%s"
-  },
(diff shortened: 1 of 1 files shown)


jpanther:
Thanks for this fix! I'm happy to merge it in.

PR #1072, opened 2025-06-22
assets: add rss and podcast icons

sudomateo:
This patch adds the Font Awesome `rss` and `podcast` icons which I believe falls under the free tier.

The icons are sourced from:
- https://fontawesome.com/icons/rss
- https://fontawesome.com/icons/podcast


jpanther:
Thanks for the PR, I'm happy to add these in.

PR #1073, opened 2025-06-22
layouts: correctly check for template existence

sudomateo:
https://github.com/jpanther/congo/commit/4e4b470915171a5855f3f2ea25851995aace8b55 fixed the theme for building with Hugo v0.146.0 or later but it doesn't correctly check for template existence. That is, if you set `homepage.layout` to `profile` this theme will still load the `page` layout.

Hugo's [templates.Exists](https://gohugo.io/functions/templates/exists/) function tests for the existence of templates relative to the `layouts` directory. The example code they provide uses `printf` to join the template name with the `_partials` directory, which would be `partials` instead for this theme.

```
{{ $partialPath := printf "headers/%s.html" .Type }}
{{ if templates.Exists ( printf "_partials/%s" $partialPath ) }}
  {{ partial $partialPath . }}
{{ else }}
  {{ partial "headers/default.html" . }}
{{ end }}
```

### DIFF
--- a/exampleSite/layouts/partials/home/custom.html
+++ b/exampleSite/layouts/partials/home/custom.html
@@ -1,9 +1,9 @@
 {{ $jsHome := resources.Get "js/home.js" | resources.Minify | resources.Fingerprint "sha512" }}
 <div id="page">
-  {{ partial "partials/home/page.html" . }}
+  {{ partial "home/page.html" . }}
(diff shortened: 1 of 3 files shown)


jpanther:
Thanks for picking this up, @sudomateo. I see where I went wrong here and I'll merge this in.

PR #1075, opened 2025-06-22
layouts: update `partials` to `_partials`

sudomateo:
Hugo v0.146.0 introduced a [new template system](https://gohugo.io/templates/new-templatesystem-overview/) that changed the structure of the `layouts` directory.

Commit https://github.com/jpanther/congo/commit/99d5cf8b28ce82377588310799420721943a3e65 updated this theme to support the new `layouts` directory structure but missed a few updates to `templates.Exists` calls. This patch fixes those calls.

### DIFF
--- a/layouts/_partials/article-link.html
+++ b/layouts/_partials/article-link.html
@@ -61,7 +61,7 @@
           {{ partial "badge.html" (i18n "article.draft" | emojify) }}
         </div>
       {{ end }}
-      {{ if templates.Exists "partials/extend-article-link.html" }}
+      {{ if templates.Exists "_partials/extend-article-link.html" }}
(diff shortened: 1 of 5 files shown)


sudomateo:
Before these changes setting `header.layout` to `hamburger` would have no effect and the default `basic` layout would be rendered.

Screenshot_2025-06-22_11-53-51

After these changes you get that sweet sweet hamburger.

Screenshot_2025-06-22_11-53-19

jpanther:
Thank you! I suspected I missed some of these. It's getting to the point where I need a whole CI suite to check all the components of this theme! I'll merge this in and will get a bug fix release out soon.

PR #1091, opened 2025-07-31
Fix code copy button when noClasses = true

angristan:
When `noClasses` is set to true, the `chroma` class is not defined on the `pre` tag. Using `pre` instead of the class to select the code block for the copy button ensures it works in both cases.

https://gohugo.io/content-management/syntax-highlighting/#noclasses

### DIFF
--- a/assets/js/code.js
+++ b/assets/js/code.js
@@ -13,7 +13,7 @@ function createCopyButton(highlightDiv) {
 }
 
 async function copyCodeToClipboard(button, highlightDiv) {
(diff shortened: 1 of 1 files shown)


angristan:
.

jpanther:
Thanks for the bug fix, @angristan. Apologies for my delayed response - I've been very busy lately. I'll merge this change in and it will be bundled in the next release.

angristan:
No worries, I can relate. Good luck with life :)

PR #1112, opened 2025-09-22
:memo: Add yandong.xyz to users showcase list

YanDongXYZ:
Thank you for creating the Congo theme — I really love it and have used it to build my personal site. I would be honored if you would add it to the users showcase list.


jpanther:
Thanks for using Congo, @YanDongXYZ! I'll merge this change in and your site will appear on the public list when the next theme release is published.

PR #1114, opened 2025-10-05
add reliable.codes to congo users page

lippoliv:
I'm working with congo now for nearly two years and I guess I've put together some interesting things that hopefully qualifies as a good candidate for the congo users list :)

Thanks for your work in this theme.


lippoliv:
I see my IDE did some auto-formatting on other rows beside mine. If this is an issue, let me know.

jpanther:
Thanks for using Congo, @lippoliv! I'll merge this change in and your site will appear on the public list when the next theme release is published.

PR #1119, opened 2025-10-20
Add blog SoftwareWitchcraft.com to user list

alxsabo:
Add blog SoftwareWitchcraft.com to user list


jpanther:
Thanks for using Congo, @alxsabo! I'll merge this change in and your site will appear on the public list when the next theme release is published.

PR #1141, opened 2026-01-26
feat: add `yashgarg.dev` in users

yash-garg:
Added my personal website to users list.


jpanther:
Thanks for using Congo, @Yash-Garg! I'll merge this change in and your site will appear on the public list when the next theme release is published.

yash-garg:
> Thanks for using Congo, @Yash-Garg! I'll merge this change in and your site will appear on the public list when the next theme release is published.

Thanks for the great theme!

PR #1146, opened 2026-02-16
📝 Add probablyalex.com to the user sites list

probablyalexzhu:
Long time user and fan of the Congo theme! Thank you


jpanther:
Thanks for using Congo, @probablyalexzhu! I'll merge this change in and your site will appear on the public list when the next theme release is published.

PR #1151, opened 2026-02-21
Declare alternate languages on translated pages

jneidel:
This MR declares alternate languages explicitly, if a page is translated.

Why is this necessary? The sitemap already declares alternates:
```html
            <xhtml:link
              rel="alternate"
              hreflang="{{ .Language.Lang }}"
              href="{{ .Permalink }}"
              />{{ end }}
```
That does work, but for ahrefs this is not enough:
image
(Feb 15 scan was after I deployed the fix on my site.)

image

After the fix:
image

Fix on `npm run example`:
image

### DIFF
--- a/layouts/_partials/head.html
+++ b/layouts/_partials/head.html
@@ -145,6 +145,13 @@
   {{ partial "vendor.html" . }}
   {{/* Analytics */}}
   {{ partial "analytics.html" . }}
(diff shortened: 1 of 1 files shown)


jpanther:
Thanks for this change. I'm happy to merge it in.

PR #1160, opened 2026-03-20
📝 Add ebourgess.dev to the user sites list

ebourgess:
Been using this theme for two years now. Thank you for making this great theme and making it opensource


jpanther:
Thanks for using Congo, @ebourgess! I'll merge this change in and your site will appear on the public list when the next theme release is published.

